PR #1, opened 2020-08-03
Implement and test single-cancer cross-validation

jjc2718:
Goal of this PR is to set up cross-validation to compare the following cases:

1. Train a mutation prediction model on 80% of data from a single cancer type, and test on the other 20% of data from that cancer type
2. Train a mutation prediction model on 80% of data from a single cancer type + all other data from TCGA, test on the other 20% of data from that cancer type

This PR just implements and tests the code to split the gene expression data. Next PR will deal with processing mutation labels and writing a script to run the comparison across the top 50 most common mutations in TCGA.

Question: currently I'm storing file paths in a config file at `pancancer_utilities/config.py`. The file paths in here work fine when I install the package in development mode (e.g. using `pip install -e .`), but when I install it the standard way (e.g. using `pip install .`) the file paths break because only the `pancancer_utilities` directory is copied and the root directory no longer exists. Do you have suggestions for a better way to handle specifying file paths and making them accessible anywhere in the root directory?

I've been trying to avoid hardcoding the root directory (currently using  `pathlib.Path(__file__)` to get it dynamically instead) but if hardcoding it is the easiest solution I'm fine with doing that.


pancancer_utilities/data_utilities.py
```diff
@@ -0,0 +1,99 @@
...
+    verbose (bool): whether or not to print verbose output
+
+    Returns
+    -------
+    rnaseq_df: samples x genes expression dataframe
+    """
+    if verbose:
+        print('Loading gene expression data...')
+
+    rnaseq_df = pd.read_csv(cfg.rnaseq_data, index_col=0, sep='\t')
```
ben-heil, line 28: Just double checking, the file is stored in samples x genes format , not genes x samples?

tests/test_cross_validation.py
```diff
@@ -0,0 +1,58 @@
...
+    assert train_df_pancancer.shape[1] == test_df_pancancer.shape[1]
+    assert train_df_pancancer.shape[1] == rnaseq_df.shape[1]
+
+    train_cancer_types = get_cancer_types(
+            sample_info_df, train_df_pancancer.index)
+    test_cancer_types = get_cancer_types(
+            sample_info_df, test_df_pancancer.index)
+
+    assert set([cancer_type]).issubset(train_cancer_types)
+    assert len(train_cancer_types.difference(set([cancer_type]))) != 0
```
ben-heil, line 56: This could be a little clearer with the code below since we already made sure that `cancer_type` is a subset of `train_cancer_types`

```suggestion
    assert len(train_cancer_types) > len(set([cancer_type])) 
```

pancancer_utilities/data_utilities.py
```diff
@@ -0,0 +1,99 @@
+import os
+
+import pandas as pd
+import pickle as pkl
+from sklearn.model_selection import KFold
+from sklearn.preprocessing import MinMaxScaler
+
+import pancancer_utilities.config as cfg
+
+def load_expression_data(subset_mad_genes=cfg.num_features_raw,
```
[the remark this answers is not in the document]
jjc2718: I'm just going to remove this argument for now. I added a comment to #2 to remind me to handle this when I implement gene filtering by MAD later on.

pancancer_utilities/data_utilities.py
jjc2718: Yep, the processed RNA-seq data (being loaded into memory here) is samples x genes. The original file from the Genome Data Commons is genes x samples, but in my data download/preprocessing script `00_download_data.ipynb` I transpose it (among other things).

For me, it's easier to store it this way since I can load it and put it straight into scikit-learn classifiers. I'm planning to subset by both genes and samples, so I don't think either orientation will be substantially better in terms of runtime (but I'll keep this in mind going forward).

pancancer_utilities/data_utilities.py
```diff
@@ -0,0 +1,99 @@
...
+    subset_mad_genes (int): TODO still need to implement this
+    scale_input (bool): whether or not to scale the expression data
+    verbose (bool): whether or not to print verbose output
+
+    Returns
+    -------
+    rnaseq_df: samples x genes expression dataframe
+    """
+    if verbose:
+        print('Loading gene expression data...')
```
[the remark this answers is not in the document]
jjc2718: I've thought about this in the past and generally stuck with stdout for log messages, since I was worried that other programs that might call my scripts would consider anything on stderr an error. But it looks like log messages on stderr is standard practice: https://unix.stackexchange.com/a/331808

I'm not sure if I'm planning to write results to stdout or directly to files, but I'll switch this to stderr for now like you suggested.

jjc2718:
Thanks for the config file/file paths ideas! I'm going to wait on this for now since I want to get my experiments running, but I created #3 to revisit how I'm doing this in the future.

pancancer_utilities/data_utilities.py
ben-heil: That's totally fair! I hadn't seen any samples x genes files before so wanted to make sure

Issue #3, opened 2020-08-03
Revisit file path configuration


jjc2718:
See Ben's suggestions [in this comment](https://github.com/greenelab/pancancer-evaluation/pull/1#pullrequestreview-460002652).

PR #6, opened 2020-08-05
Add script to run pancancer classification experiments

jjc2718:
This PR adds the main script to run pan-cancer or single-cancer classification for a given gene and cancer type. This is implemented in `pancancer_utilities/scripts/classify_cancer_type.py`. Other functions that are being added (i.e. the ones in `pancancer_utilities/tcga_utilities.py` and `pancancer_utilities/classify_utilities.py`) are mostly copies of code from Greg's BioBombe and pancancer repos, so these don't need to be reviewed super closely.

A script to run this over multiple genes and cancer types is included at `01_run_pancancer_classification.py`. For now, this is mostly a proof of concept to make sure things work; the results of this particular experiment (in `02_plot_results.ipynb`) aren't very interesting so I haven't spent much time documenting them yet.

In future PRs I plan to run these experiments for more genes and cancer types, and to implement statistical testing to identify genes that benefit from including pan-cancer data.


pancancer_utilities/tcga_utilities.py
```diff
@@ -0,0 +1,180 @@
...
+    Returns
+    -------
+    Write file of cancer-type filtering to disk and output a processed y vector
+    """
+    if include_copy:
+        y_df = y_copy + y_mutation
+    else:
+        y_df = y_mutation
+
+    y_df.loc[y_df > 1] = 1
```
ajlee21, line 49: I'm not very familiar with this type of data, so maybe this is obvious. But why are you setting a cap to 1?

pancancer_utilities/tcga_utilities.py
```diff
@@ -0,0 +1,180 @@
...
+    y_df = pd.DataFrame(y_df)
+    y_df.columns = ["status"]
+
+    y_df = (
```
ajlee21, line 53: Might be a little lost here...so `y_df` is a df with a single `status` column? Does that mean that mutation burden and copy status values are stacked on top of each other? And I assume the rows are sample ids. If there is only one column doesn't that means that you'll have duplicate sample ids since I assume samples will have mutation and copy status info associated?

pancancer_utilities/tcga_utilities.py
```diff
@@ -0,0 +1,180 @@
...
+    Combine copy number and mutation data and filter cancer-types to build y matrix
+
+    Arguments
+    ---------
+    y_mutation: Pandas DataFrame of mutation status
+    y_copy: Pandas DataFrame of copy number status
+    include_copy: boolean if the copy number data should be included in status calc
+    gene: string indicating gene of interest (used for writing proportion file)
+    sample_freeze: pandas dataframe storing which samples to use
+    mutation_burden: pandas dataframe storing log10 mutation counts
```
ajlee21, line 34: Just checking my understanding. So mutation burden and copy status are perform at a sample-level where mutation burden indicates how many mutations compared to some reference and copy status is how many copies of a gene respectively?

pancancer_utilities/tcga_utilities.py
```diff
@@ -0,0 +1,180 @@
...
+    y_df = (
+        y_df.merge(
+            sample_freeze, how="left", left_index=True, right_on="SAMPLE_BARCODE"
+        )
+        .set_index("SAMPLE_BARCODE")
+        .merge(mutation_burden, left_index=True, right_index=True)
+    )
+
+    # Get statistics per gene and disease
+    disease_counts_df = pd.DataFrame(y_df.groupby("DISEASE").sum()["status"])
```
ajlee21, line 62: Ok based on this I think I might be confused as to the format of `y_df`

pancancer_utilities/scripts/classify_cancer_type.py
```diff
@@ -0,0 +1,244 @@
...
+include_copy = True
+if classification == "Oncogene":
+    y_copy_number_df = copy_gain_df.loc[:, gene_name]
+elif classification == "TSG":
+    y_copy_number_df = copy_loss_df.loc[:, gene_name]
+else:
+    y_copy_number_df = pd.DataFrame()
+    include_copy = False
+
+y_df = process_y_matrix(
```
ajlee21, line 115: Might be helpful to add a comment here on what this is returning. Or in the function itself. I had a little trouble following..but I'm also outside of my field on this one

01_run_pancancer_classification.py
```diff
@@ -0,0 +1,49 @@
...
+    if use_pancancer:
+        args.append('--use_pancancer')
+    if shuffle_labels:
+        args.append('--shuffle_labels')
+
+    if verbose:
+        print('Running: {}'.format(' '.join(args)))
+    subprocess.call(args)
+
+if __name__ == '__main__':
```
ajlee21, line 41: Still new to pipeline building. This seems interesting to me, I wonder if you could walk me through the logic here. So you have this main function that calls this `run_single_experiment`, which generates a python terminal call...is there a reason to have it as a terminal call?

01_run_pancancer_classification.py
```diff
@@ -0,0 +1,49 @@
...
+import pancancer_utilities.config as cfg
+
+EXP_SCRIPT = os.path.join(cfg.repo_root,
+                          'pancancer_utilities',
+                          'scripts',
+                          'classify_cancer_type.py')
+
+# genes and cancer types to run experiments for
+# just hardcoding these for now, might choose them systematically later
+genes = ['TP53', 'PTEN', 'KRAS', 'BRAF', 'TTN']
```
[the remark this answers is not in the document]
jjc2718: Yes, I definitely plan to come up with a better/more general way to do this in the future. What I have now is mostly just to make sure things run and give (roughly) the output I would expect.

01_run_pancancer_classification.py
```diff
@@ -0,0 +1,49 @@
...
+with the corresponding arguments
+"""
+import os
+import sys
+import subprocess
+
+import pandas as pd
+from tqdm import tqdm
+
+import pancancer_utilities.config as cfg
```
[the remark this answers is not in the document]
jjc2718: I don't love the way I'm handling file paths right now, so I'm not sure if I'd recommend doing it exactly this way. In my last PR (#1) Ben had some alternative suggestions that I liked, so I made #3 to come back to it later.

If you're interested (and for my own reference in the future), the problem with what I have now is that it relies on the directory being set up exactly like it currently is. If you install the `pancancer_utilities` package using `pip install -e .`, the `pancancer_utilities` import uses the local version of the code and everything is fine. But if you install it in non-development mode using `pip install .`, everything in the `pancancer_utilities` directory gets copied somewhere else which breaks the file paths.

So what I'm doing works if you make it very clear in the README that `pip install -e .` is the right way to install the utilities functions, but it seems to me to not really be the intended use of pip development mode. I'd like to find a more general way to handle file paths that doesn't rely on that. I'll let you know what I end up deciding on; happy to also chat offline if you want.

pancancer_utilities/scripts/classify_cancer_type.py
```diff
@@ -0,0 +1,244 @@
...
+gene_aupr_list = []
+gene_coef_list = []
+gene_metrics_list = []
+
+# create directory for the gene
+dirname = 'pancancer' if args.use_pancancer else 'single_cancer'
+gene_dir = os.path.join(args.results_dir, dirname, gene_name)
+os.makedirs(gene_dir, exist_ok=True)
+
+# check if gene has been processed already
```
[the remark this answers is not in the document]
jjc2718: Will edit this to clarify (hopefully). I just meant the experiment has been run already so it can be skipped.

pancancer_utilities/tcga_utilities.py
jjc2718: At this point, `y_df` contains information about how many deleterious mutations or copy number alterations were observed in the given gene for the given sample. We want to make this into a classification problem (either a sample has or doesn't have a mutation), so anything with 1 or more mutations gets a label of 1.

It would be interesting to treat this instead as an ordinal regression problem, i.e. predicting the number of mutations/CNVs from gene expression, but this might get a bit complicated since there are lots of subtle biases that go into the number of mutations (background mutation rates vary a lot across cancers and across samples, etc) and interpreting the results would be harder. Interesting future question, though!

pancancer_utilities/tcga_utilities.py
jjc2718: > so y_df is a df with a single status column? Does that mean that mutation burden and copy status values are stacked on top of each other?

Yep, exactly. We're defining a "mutation" to be any deleterious point mutation *or* copy number change in the given gene. So the line at the top of the function `y_df = y_copy + y_mutation` is taking the sum of CNVs and point mutations in the gene.

> doesn't that means that you'll have duplicate sample ids since I assume samples will have mutation and copy status info associated?

I'll admit that I just copied this line from Greg's code, but I think taking the sum of dataframes that have the same indexes (sample IDs in this case) should just take the sum of each of the entries. The list of sample IDs should be the same, then, after taking the sum. I'll check this by running some simple examples in a notebook.

pancancer_utilities/tcga_utilities.py
jjc2718: Yep, exactly!

pancancer_utilities/tcga_utilities.py
jjc2718: This is another holdover from Greg's code, but I think the merge calls above add a column for the cancer type of each sample (called `DISEASE`), then the code below filters cancer types (diseases) based on the proportion of positively labeled samples. It doesn't make sense to do classification if a gene isn't mutated in any samples (or is mutated in almost every sample), so we just ignore those cancer types. 

Let me know if you still have questions - they're useful for helping me to understand this code better too!

01_run_pancancer_classification.py
jjc2718: Not a good reason, other than that a lot of this code existed as a script before so it was easiest. I think longer-term it would be best to refactor this file to be a series of functions, which I could then just import into a script like this (rather than having to use `subprocess`).

This would also have the advantage that I'd only have to load all of the data once, as opposed to loading it each time I call the script. I have #7 to remind me to make this change in the future.

jjc2718:
Answers to your questions:

> So is the goal to predict mutation burden and copy number variation (label) for a given sample from a specific cancer type A. Where you're training set is either only samples from cancer A or samples from all types of cancer? 

Yep, that's exactly right! And we're combining point mutation and CNV data like I described in my comment above. I haven't really done the actual comparison yet (eventually we want to do statistical testing for which is better, and things like that) but this PR is mostly just the setup of the core experiment scripts.

> Also what the input data in this case? expression data?

Yes! In this case I'm just using expression data. In the future we'll also include features for sample mutation burden and cancer type (those have helped in Greg's work), but for now it's just TCGA gene expression.

PR #10, opened 2020-08-10
Subset gene features by MAD and clean up data preprocessing code

### LINKED ISSUE #2: Implement subsetting genes by MAD
See TODOs in `pancancer_utilities/data_utilities.py`.

This will probably end up looking a bit like [the code here](https://github.com/greenelab/BioBombe/blob/8ebb47c6f3b736a589c705ea51fcb1f70e856f87/0.expression-download/scripts/nbconverted/2A.process-pancanatlas-data.py#L260) and [here](https://github.com/greenelab/BioBombe/blob/8ebb47c6f3b736a589c705ea51fcb1f70e856f87/9.tcga-classify/classify-with-raw-expression.py#L82), but will need to be modified slightly since I'm doing the cross-validation stratification a bit differently here.

### LINKED ISSUE #4: Cross-validation for imbalanced label case
If labels are highly imbalanced (for example, TP53 in ovarian cancer) ROC can break because some cross-validation splits will only have one class.

Maybe using [StratifiedKFold](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.StratifiedKFold.html) instead of standard k-fold CV is the best solution here?

### LINKED ISSUE #5: Fix cancer type covariate when single cancer is held out
See code in `pancancer_utilities/tcga_utilities.py`, specifically the `align_matrices` function. This won't work if the test set has a different cancer type composition than the training set, so there needs to be a way to preserve this across datasets.

jjc2718:
This PR does the following:
* Implements the preprocessing step of subsetting genes by mean absolute deviation (MAD) (closes #2)
* Splits code for standardizing features and for filtering samples, previously this happened in the same function (closes #5)

See [README.md](https://github.com/greenelab/pancancer-evaluation/blob/master/README.md) for a description of the project. Happy to discuss/answer questions/add clarification if what I'm doing here is unclear.

### DIFF
--- a/pancancer_utilities/data_utilities.py
+++ b/pancancer_utilities/data_utilities.py
@@ -92,7 +92,7 @@ def load_pancancer_data(gene_list, verbose=False):
         with open(cfg.pancan_data, 'wb') as f:
             pkl.dump(pancan_data, f)
 
-    return (genes_df, pancan_data)
+    return genes_df, pancan_data
 
 def load_top_50():
     """Load top 50 mutated genes in TCGA from BioBombe repo.
@@ -185,7 +185,7 @@ def split_by_cancer_type(rnaseq_df, sample_info_df, holdout_cancer_type,
     else:
         rnaseq_train_df = cancer_type_train_df
 
-    return (rnaseq_train_df, rnaseq_test_df)
+    return rnaseq_train_df, rnaseq_test_df
 
 def split_single_cancer_type(cancer_type_df, num_folds, fold_no, seed):
     """Split data for a single cancer type into train and test sets."""
@@ -267,9 +267,12 @@ def subset_by_mad(X_train_df, X_test_df, gene_features, subset_mad_genes, verbos
     if verbose:
         print('Taking subset of gene features', file=sys.stderr)
 
-    mad_genes_df = pd.DataFrame(
-            X_train_df.loc[:, gene_features].mad(axis=0).sort_values(ascending=False)
-    ).reset_index()
+    mad_genes_df = (
+        X_train_df.loc[:, gene_features]
+                  .mad(axis=0)
+                  .sort_values(ascending=False)
+                  .reset_index()
+    )
     mad_genes_df.columns = ['gene_id', 'mean_absolute_deviation']
     mad_genes = mad_genes_df.iloc[:subset_mad_genes, :].gene_id.astype(str).values
 
@@ -282,6 +285,6 @@ def subset_by_mad(X_train_df, X_test_df, gene_features, subset_mad_genes, verbos
     ))
     train_df = X_train_df.reindex(valid_features, axis='columns')
     test_df = X_test_df.reindex(valid_features, axis='columns')
-    return (train_df, test_df, gene_features)
+    return train_df, test_df, gene_features
 
 
--- a/pancancer_utilities/data_utilities.py
+++ b/pancancer_utilities/data_utilities.py
@@ -5,6 +5,7 @@ Functions for reading/writing/processing data
 import os
 import sys
 
+import numpy as np
 import pandas as pd
 import pickle as pkl
 from sklearn.model_selection import KFold
@@ -245,7 +246,7 @@ def summarize_results(results, gene, holdout_cancer_type, signal, z_dim,
 
     return metrics_out_, roc_df_, pr_df_
 
-def subset_by_mad(X_train_df, X_test_df, subset_mad_genes, verbose=False):
+def subset_by_mad(X_train_df, X_test_df, gene_features, subset_mad_genes, verbose=False):
     """Subset features by mean absolute deviation.
 
     Takes the top subset_mad_genes genes (sorted in descending order),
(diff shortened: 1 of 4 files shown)


pancancer_utilities/data_utilities.py
```diff
@@ -252,4 +247,41 @@ def summarize_results(results, gene, holdout_cancer_type, signal, z_dim,
...
+    """Subset features by mean absolute deviation.
+
+    Takes the top subset_mad_genes genes (sorted in descending order),
+    calculated on the training set.
+
+    Arguments
+    ---------
+    X_train_df: training data, samples x genes
+    X_test_df: test data, samples x genes
+    gene_features: numpy array of gene features
```
miltondp, line 260: Looks like `gene_features` is an array of bool, right? It might be good to add that here.

I understand you have different features in `X_train_df`, and this parameter indicates which ones are genes.

pancancer_utilities/data_utilities.py
```diff
@@ -252,4 +247,41 @@ def summarize_results(results, gene, holdout_cancer_type, signal, z_dim,
...
+    Returns
+    -------
+    (train_df, test_df, gene_features) datasets with filtered features
+    """
+    if verbose:
+        print('Taking subset of gene features', file=sys.stderr)
+
+    mad_genes_df = pd.DataFrame(
+            X_train_df.loc[:, gene_features].mad(axis=0).sort_values(ascending=False)
+    ).reset_index()
```
miltondp, line 272: It should not be necessary to create another DataFrame I think.
```suggestion
mad_genes_df = X_train_df.loc[:, gene_features].mad(axis=0).sort_values(ascending=False).reset_index()
```

pancancer_utilities/data_utilities.py
```diff
@@ -252,4 +247,41 @@ def summarize_results(results, gene, holdout_cancer_type, signal, z_dim,
...
+    non_gene_features = X_train_df.columns.values[~gene_features]
+    valid_features = np.concatenate((mad_genes, non_gene_features))
+
+    gene_features = np.concatenate((
+        np.ones(mad_genes.shape[0]).astype('bool'),
+        np.zeros(non_gene_features.shape[0]).astype('bool')
+    ))
+    train_df = X_train_df.reindex(valid_features, axis='columns')
+    test_df = X_test_df.reindex(valid_features, axis='columns')
+    return (train_df, test_df, gene_features)
```
miltondp, line 285: Another minor. You are using Python 3.6, but I guess you can remove parenthesis here:
```suggestion
    return train_df, test_df, gene_features
```

pancancer_utilities/data_utilities.py
```diff
@@ -5,6 +5,7 @@
 import os
```
miltondp, line 5: A very minor suggestions: according to PEP 8 E302, two blank lines should be left between methods.

pancancer_utilities/scripts/classify_cancer_type.py
```diff
@@ -18,6 +18,7 @@
     process_y_matrix,
     process_y_matrix_cancertype,
```
miltondp, line 19: `process_y_matrix_cancertype` is unused in this file, as well as `import pickle as pkl`

pancancer_utilities/scripts/classify_cancer_type.py
```diff
@@ -41,13 +46,34 @@
...
 p.add_argument('--verbose', action='store_true')
 args = p.parse_args()
 
 np.random.seed(args.seed)
 if args.verbose:
     logging.basicConfig(level=logging.DEBUG, format='%(message)s')
 
+# create directory for the gene
+dirname = 'pancancer' if args.use_pancancer else 'single_cancer'
+gene_dir = os.path.join(args.results_dir, dirname, args.gene)
```
miltondp, line 61: Another minor suggestion. Thanks to @danich1 I now use `pathlib` a lot.
```suggestion
gene_dir = Path(args.results_dir, dirname, args.gene).resolve()
```

pancancer_utilities/scripts/classify_cancer_type.py
```diff
@@ -41,13 +46,34 @@
...
+os.makedirs(gene_dir, exist_ok=True)
```
miltondp, line 62: And this would be:
```suggestion
gene_dir.mkdir(parents=True, exist_ok=True)
```

pancancer_utilities/scripts/classify_cancer_type.py
```diff
@@ -128,32 +144,54 @@
...
+        X_train_raw_df, X_test_raw_df = du.split_by_cancer_type(
+           rnaseq_df, sample_info_df, args.holdout_cancer_type,
+           num_folds=args.num_folds, fold_no=fold_no,
+           use_pancancer=args.use_pancancer, seed=args.seed)
     except ValueError:
         exit('No test samples found for cancer type: {}, gene: {}\n'.format(
                args.holdout_cancer_type, args.gene))
 
+    y_train_df = y_df.reindex(X_train_raw_df.index)
+    y_test_df = y_df.reindex(X_test_raw_df.index)
```
miltondp, line 177: why `y_df` here is the same for both training and testing sets?

pancancer_utilities/scripts/classify_cancer_type.py
```diff
@@ -128,32 +144,54 @@
...
+
+    # data processing/feature selection, needs to happen for train and
+    # test sets independently
```
miltondp, line 180: Yes, nice thread you retweeted the other day https://twitter.com/jjc2718/status/1291182838612660224

pancancer_utilities/tcga_utilities.py
```diff
@@ -120,48 +121,71 @@ def process_y_matrix_cancertype(
...
     """
     Process the x matrix for the given input file and align x and y together
 
     Arguments
     ---------
     x_file_or_df: string location of the x matrix or matrix df itself
     y: pandas DataFrame storing status of corresponding samples
-    algorithm: a string indicating which algorithm to subset the z matrices
+    add_cancertype_covariate: if true, add one-hot encoded cancer type as a covariate
+    add_mutation_covariate: if true, add log10(mutation burden) as a covariate
```
miltondp, line 134: If these `add_*` methods keep increasing, you might add a list with functions to generically modify the data.

pancancer_utilities/tcga_utilities.py
```diff
@@ -120,48 +121,71 @@ def process_y_matrix_cancertype(
...
+    Note this should be done for train and test sets independently. Also note
+    this doesn't necessarily preserve the order of features (this shouldn't
+    matter in most cases).
+    """
+    x_df_gene = x_df.loc[:, gene_features]
+    x_df_other = x_df.loc[:, ~gene_features]
+    x_df_scaled = pd.DataFrame(
+        StandardScaler().fit_transform(x_df_gene),
+        index=x_df_gene.index,
+        columns=x_df_gene.columns
```
miltondp, line 186: I remember having issues in the past when using the same object instance for index/columns of another DataFrame in a new one. I always use this just in case:
```suggestion
        index=x_df_gene.index.copy(),
        columns=x_df_gene.columns.copy()
```

pancancer_utilities/data_utilities.py
```diff
@@ -252,4 +247,41 @@ def summarize_results(results, gene, holdout_cancer_type, signal, z_dim,
...
+    X_train_df: training data, samples x genes
+    X_test_df: test data, samples x genes
+    gene_features: numpy array of gene features
+    subset_mad_genes (int): number of genes to take
+
+    Returns
+    -------
+    (train_df, test_df, gene_features) datasets with filtered features
+    """
+    if verbose:
```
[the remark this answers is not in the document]
jjc2718: Good idea! Created #11 to add this in the future.

pancancer_utilities/scripts/classify_cancer_type.py
jjc2718: Good catch, I copied those over from another script but ended up not using them.

pancancer_utilities/tcga_utilities.py
jjc2718: Good idea! I'll keep that in mind.

pancancer_utilities/scripts/classify_cancer_type.py
jjc2718: Essentially, the process is as follows:

- generate labels (stored in `y_df`)
- split data into train and test sets (these are `X_train_raw_df` and `X_test_raw_df`)
- split labels into train and test sets (so these lines split `y_df` into `y_train_df` and `y_test_df`)

So the training and testing labels come from a partition of all of the labels, which is what these lines are doing. Let me know if that's still not clear.

miltondp: It's clear, thank you, Jake!

PR #14, opened 2020-08-24
Refactor data loading code to load data once for multiple experiments

### LINKED ISSUE #7: Refactor classification script so data only gets loaded once
Currently, experiments work by calling `classify_cancer_type.py` for each gene/cancer type combination. This script reloads all of the data into memory each time, which takes 5-10 minutes.

Creating classifiers for many genes/cancer types would be a lot faster if the code were refactored to load the data once, then run all combinations subsequently (i.e in a single script).

jjc2718:
Closes #7. Changes in this PR:
* Moved mutation prediction script (previously `classify_cancer_type.py`) to a class, allowing data to be loaded once and kept in memory for all subsequent experiments.
* Get rid of `subprocess` calls in `01_run_pancancer_classification.py`, changing it to initialize the class once and preprocess data only as needed for each experiment.
* Add debugging option to only load a small subset of the data, add option to load data from Vogelstein paper (see `data_utilities.py`), and some other small changes

This refactor will allow me to run multiple experiments much faster, since I don't have to reload data and make `subprocess` calls each time. Sorry for the fairly large PR, no rush to get it reviewed.

Historically I haven't done too much OOP (my view is basically [this tweet](https://twitter.com/chrisalbon/status/1296597637013966851?s=20)) so any feedback on how I could structure this better or make it more object-oriented would be awesome.


pancancer_evaluation/mutation_prediction.py
```diff
@@ -0,0 +1,465 @@
...
+        if shuffle_labels:
+            y_filtered_df.status = np.random.permutation(
+                y_filtered_df.status.values)
+
+        self.X_df = rnaseq_filtered_df
+        self.y_df = y_filtered_df
+        self.gene_features = gene_features
+
+
+    def run_cv_for_cancer_type(self, gene, cancer_type, sample_info,
```
ben-heil, line 133: Is it necessary to couple training models with data processing? If you decide you want to try two different ways of processing data down the line you'll have to either make another class or modify the training logic. If your cv function is separate from data processing and takes in a list of dataframes as an argument, then you can use it with any kind of preprocessing without needing to modify it

ben-heil: To clarify, I agree that it makes sense to only load the data once, but you can load the data with this class and then pass it into separate training functions unless I'm missing something

pancancer_evaluation/mutation_prediction.py
```diff
@@ -0,0 +1,465 @@
...
+                    'gene: {}\n'.format(cancer_type, gene)
+                )
+
+            self.results['gene_metrics'].append(metric_df)
+            self.results['gene_auc'].append(gene_auc_df)
+            self.results['gene_aupr'].append(gene_aupr_df)
+            self.results['gene_coef'].append(coef_df)
+
+        self._save_results(gene, cancer_type, signal)
+        del self.results
```
ben-heil, line 251: If you delete this variable at the end of the function shouldn't it be a local variable instead of a class member variable?

pancancer_evaluation/mutation_prediction.py
```diff
@@ -0,0 +1,465 @@
...
+
+
+    def _get_metrics(self,
```
ben-heil, line 254: This is a great idea for a function and I need to add something like this to my code

pancancer_evaluation/mutation_prediction.py
jjc2718: That makes more sense, good catch

pancancer_evaluation/mutation_prediction.py
jjc2718: What you're saying makes sense to me, I guess I'm just having trouble picturing how it would work. So model training (currently in this function, `run_cv_for_cancer_type`) would be a separate class or function in a separate file, that just takes the data (self.X and self.y) from this class as arguments?

Maybe an example would be helpful - do you know of any examples that do what you're describing (either from your own code or from other repos)?

ben-heil: Right, the function currently only uses the seed, `X_df`, and `y_df` from the class it lives in, along with a couple functions. You could write the same function with `X_df` and `y_df` as parameters and it wouldn't have to live in the class. You would then have to call `preprocess_data`, `save_results` etc from outside the class.

I guess the idea is that the concept of training and the concept of data management are two different things that don't really need to know about each other. The code as it currently is looks fine, it just might be harder to manage in the future since changes to data processing now might affect training

jjc2718: Makes sense. I'll split this into another PR since it will be a decent sized change, but I agree it would be better to separate these two steps.

PR #20, opened 2020-09-11
Stratified cross-validation experiments/visualizations

jjc2718:
This implements a cross-validation scheme similar to what's used in the BioBombe paper. This will provide a baseline for comparing the results when single cancer types are held out, so that we know what works when all cancer types are held out (we would expect to see some overlap here, but the differences will be interesting).

I tried this for two gene sets: the top 50 most mutated genes in TCGA (this is what BioBombe uses), and a set of known oncogenes/tumor suppressors from [this paper](https://doi.org/10.1038/nm1087) (data on oncogene/TSG status comes from here: https://github.com/greenelab/pancancer/blob/master/data/vogelstein_cancergenes.tsv).

Also added statistical testing to compare results against our negative control (shuffling the true mutation labels), and added some code in `4_plot_results.ipynb` to visualize the results.


pancancer_evaluation/utilities/analysis_utilities.py
```diff
@@ -0,0 +1,79 @@
...
+        signal_results = results_df[
+            (results_df[identifier] == id_str) &
+            (results_df.data_type == 'test') &
+            (results_df.signal == 'signal')
+        ][metric].values
+        shuffled_results = results_df[
+            (results_df[identifier] == id_str) &
+            (results_df.data_type == 'test') &
+            (results_df.signal == 'shuffled')
+        ][metric].values
```
ben-heil, line 60: You can avoid having a hanging `[metric].values` by declaring the conditions ahead of time like in the code below. It's just a style choice, but seeing a right bracket on a newline confused me for a second. (If you decide to use the suggestion, double check it. I always screw up multiple conditions in pandas the first time)

```suggestion
    conditions = ((results_df[identifier] == id_str) &
        (results_df.data_type == 'test') &
        (results_df.signal == 'signal')
        )
    signal_results = results_df[conditions][metric].values
    
    conditions = ((results_df[identifier] == id_str) &
         (results_df.data_type == 'test') &
         (results_df.signal == 'shuffled')
         )
    shuffled_results = results_df[conditions][metric].values
```

tests/test_cross_validation.py
```diff
@@ -53,5 +56,40 @@ def test_cv_pancancer(expression_data, cancer_type):
...
+    # check that proportions of each stratification are approximately
+    # the same between folds (in other words, check that stratified CV is
+    # working properly)
+    #
+    # note that the absolute scale of the proportions can vary quite a bit,
+    # but on a relative scale they should be pretty close
+    for ix1, ix2 in it.combinations(range(num_folds), 2):
+        assert np.allclose(train_proportions[ix1], train_proportions[ix2], rtol=1.0)
+        assert np.allclose(test_proportions[ix1], test_proportions[ix2], rtol=1.0)
+        assert np.allclose(train_proportions[ix1], test_proportions[ix2], rtol=1.0)
```
ben-heil, line 95: Since `np.allclose()` isn't symmetric, it may make sense to use `it.permutations` here, but the edge case where the test passes because the second index's values are large seems unlikely if you have more than ~2 cancer types

ben-heil (approved):
Your plots look great! I recognize some of the genes in the top right hahaha

Stratifying by cancer type certainly makes sense in terms of machine learning. Will you also stratify by cancer type in your pan-cancer classifier? Do you think modulating the cancer proportions you use based on mutation signature similarity to your target cancer would help?

pancancer_evaluation/utilities/analysis_utilities.py
jjc2718: Good idea! I like your approach better, it's definitely cleaner IMO.

tests/test_cross_validation.py
jjc2718: Good call, I'll change it

jjc2718:
Responses to your questions:

> Stratifying by cancer type certainly makes sense in terms of machine learning. Will you also stratify by cancer type in your pan-cancer classifier?

IMO, the stratification that I'm doing here is mostly useful in terms of evaluating the model (i.e. making the test set "look like" the training set). In the current experiments I'm interested in how this differs when the test set is only a single cancer type (forthcoming PRs), as opposed to an equal mix of cancer types like I have here.

I think it would also be interesting to see how different proportions of each cancer type in the training set affects model performance (which I think is what you're alluding to with your questions). For instance, right now some cancer types have many more samples than others, and you could imagine trying to balance these proportions in the training set in various ways. I think this is a slightly different question that I'm not planning to address in the current set of experiments, but I'm definitely interested in it for the future!

> Do you think modulating the cancer proportions you use based on mutation signature similarity to your target cancer would help?

This is a neat idea! I'll definitely keep this in mind as well.

PR #22, opened 2020-09-18
Fix cross-validation order bug

jjc2718:
Recently I noticed that running cross-validation multiple times using the same seed gives different results. This fixes that by switching from set intersections to Pandas built-in index intersections (the latter guarantees the same order every time). I've checked that this now leads to repeatable cross-validation results.

I'm also planning to add some regression tests to make sure this isn't happening in the future, but those will be part of my next (larger) PR.

### DIFF
--- a/pancancer_evaluation/utilities/data_utilities.py
+++ b/pancancer_evaluation/utilities/data_utilities.py
@@ -209,7 +209,7 @@ def split_stratified(rnaseq_df, sample_info_df, num_folds=4, fold_no=1,
     ] = 'other'
 
     # now do stratified CV splitting and return the desired fold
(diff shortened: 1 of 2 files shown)


ben-heil (approved):
The code looks good.

Isn't set intersection an operation with a unique output? Is the issue that because you're casting sets to lists later on the order of the resulting index isn't deterministic?

jjc2718:
> Isn't set intersection an operation with a unique output? Is the issue that because you're casting sets to lists later on the order of the resulting index isn't deterministic?

Yeah, I'm pretty sure casting the sets to lists is causing the problem. Even the `reindex` calls directly after the lines I changed were returning different output each time - I assume Pandas internally casts the sets to lists, then reorders the resulting dataframe using the list order.

PR #24, opened 2020-09-23
Separate data loading from model training + add model tests

### LINKED ISSUE #7: Refactor classification script so data only gets loaded once

jjc2718:
Closes #7. Sorry this PR is a bit large, but I couldn't find any good way to break this up: most of the changes are dependent on one another.

This PR has 2 main goals:
* Decouple code for loading and preprocessing data from code used to train the model (see #7).
* Once that's done, write regression tests to make sure that model output doesn't change when I make future modifications to the code (tests run on a small subset of the data, to keep them running fairly quickly).

This should make it slightly easier if we want to try different datasets or different models in the future.


tests/test_model.py
```diff
@@ -0,0 +1,56 @@
...
+                                    check_gene_file=True,
+                                    shuffle_labels=False)
+    results = cu.run_cv_cancer_type(tcga_data, gene, cancer_type,
+                                    sample_info_df, num_folds=4,
+                                    use_pancancer=False,
+                                    shuffle_labels=False)
+    metrics_df = pd.concat(results['gene_metrics'])
+    results_file = cfg.test_cancer_type_results.format(gene, cancer_type)
+    old_results = np.loadtxt(results_file)
+    assert np.allclose(metrics_df['auroc'].values, old_results)
```
ben-heil, line 55: Writing tests to make sure the models run the same way each time is a good idea! I wish I had thought of that before my most recent issues with nondeterminism

ben-heil (approved):
Looks good to me!

tests/test_model.py
jjc2718: Yeah, I've definitely learned this the hard way in the past - hopefully it will help me catch issues in the future.

pancancer_evaluation/data_models/tcga_data_model.py
```diff
@@ -0,0 +1,234 @@
...
+        subset_mad_genes (int): how many genes to keep (top by mean absolute deviation).
+                                -1 doesn't do any filtering (all genes will be kept).
+        verbose (bool): whether or not to write verbose output
+        debug (bool): if True, use a subset of expression data for quick debugging
+        test (bool): if True, don't save results to files
+        """
+        # save relevant parameters
+        np.random.seed(seed)
+        self.seed = seed
+        self.results_dir = results_dir
```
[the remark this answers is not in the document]
jjc2718: Good point. I put it here because doing the file handling in the scripts that call this class results in a lot of boilerplate that's pretty similar between scripts, but I agree that the data model probably isn't the best place for it.

I'll move the repeated file handling code to a separate utilities file, and call that from my experiment scripts - that way it's separate from this data model, which makes more sense to me as well.

PR #29, opened 2020-10-06
Add script to analyze coefficients of model fits

jjc2718:
This analysis answers the following questions about the coefficients of mutation prediction models fit to pan-cancer data:
* How often are the one-hot encoded cancer types used by the elastic net models (i.e. left as nonzero after fitting the model)?
* How often is mutation burden used (left as nonzero) by the models?
* Which genes are used by the models? Are these mostly the genes that vary the most (i.e. highest mean absolute deviation), or are some of them less variable genes?

For all of these questions, we segment coefficients by well-performing (statistically better than the negative control with shuffled labels) and poor-performing (statistically the same or worse than the negative control).


06_coefficient_analysis.ipynb
```diff
@@ -0,0 +1,848 @@
...
+    }
+   ],
+   "source": [
+    "mad_merge_df = (\n",
+    "    mad_ids_df.merge(pancancer_comparison_df, left_on='identifier', right_on='identifier')\n",
+    "              .drop(columns=['delta_mean', 'p_value', 'corr_pval', 'nlog10_p'])\n",
+    ")\n",
+    "sns.set({'figure.figsize': (8, 6)})\n",
+    "sns.histplot(data=mad_merge_df, x='mad_proportion', hue='reject_null', kde=False)\n",
+    "plt.xlabel('Proportion of selected/nonzero genes in top {} MAD genes'.format(n_mad_genes))\n",
```
ben-heil, line 822: The takeaway that I get from this plot is that the distribution of variances in models are roughly the same for high performing and low performing models. Is that your take as well?

Does that result still hold true if you multiple-test correct? Do you think that the p-values are uniformly distributed across these bins, or is there a region where the extremely high performing models live?

ben-heil (approved):
Good work!

06_coefficient_analysis.ipynb
```diff
@@ -0,0 +1,848 @@
...
+    "import pancancer_evaluation.utilities.analysis_utilities as au"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "results_dir = '/home/jake/data/pancancer_vogelstein_09_25'\n",
```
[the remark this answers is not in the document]
jjc2718: Not a good reason - I just did this because I'm running some other experiments, and didn't want them to overwrite my previous results. I'll get rid of this.

06_coefficient_analysis.ipynb
jjc2718: > The takeaway that I get from this plot is that the distribution of variances in models are roughly the same for high performing and low performing models. Is that your take as well?

Yep, although I'd phrase it slightly differently. This plot shows (to me) that the proportion of model signal captured by the most variable genes in the dataset: (1) is roughly the same for high-performing and low-performing models, and (2) is not all that high, with a mean around 0.5 of selected genes in the model being in the set of most variable genes.

My (mostly unfounded) hypothesis was that most genes selected by the model would be highly variable genes, which would permit us to just use the most variable genes and get rid of the others. That doesn't seem to be the case, though.

> Does that result still hold true if you multiple-test correct?

To classify models into high-performing and low-performing, I'm using t-tests (comparing each model with one with shuffled labels), with FDR correction for the resulting p-values. I could see what happens at different significance cutoffs, though, which I think is what you're getting at here:

> Do you think that the p-values are uniformly distributed across these bins, or is there a region where the extremely high performing models live?

This is an interesting question - it is possible that I'm losing some information here by choosing a cutoff and binarizing into significant and not significant, and if I look at how the actual p-value corresponds to these results I might see more of a pattern. I'm not too interested in diving deeper into this at the moment (I think this result is enough for me to say that we need to keep using all the genes in these models, at least for now), but I'll keep this idea in mind for the future.

PR #30, opened 2020-10-08
Remove relevant cancer type experiments + analysis

jjc2718:
As a reminder of what I've been doing previously, I'm trying to predict cancer mutations from gene expression in individual cancer types, either training my models on data from the same cancer type or the same cancer type + all other cancers in TCGA.

As a sanity check, we wanted to try an experiment where we remove the relevant cancer type from the pan-cancer data (i.e. train on all other cancers besides X and test on X). We assume that this will always perform worse than a training set with the relevant cancer type (i.e. our pan-cancer training set from before).

Thankfully, this does seem to be the case (see the second volcano plot in `06_plot_remove_cancer_type.ipynb`, copied here: negative x-axis values = better performance with relevant cancer type).

image


06_plot_remove_cancer_results.ipynb
```diff
@@ -0,0 +1,1534 @@
...
+       "        vertical-align: middle;\n",
+       "    }\n",
+       "\n",
+       "    .dataframe tbody tr th {\n",
+       "        vertical-align: top;\n",
+       "    }\n",
+       "\n",
+       "    .dataframe thead th {\n",
+       "        text-align: right;\n",
+       "    }\n",
```
ajlee21, line 590: Nice plot!

So it looks like you have some genes that very accurately predict mutation status using pancancer but not using pancancer-only. And these genes are not signficantly more accurate compared to your shuffled data. What would you say about these genes? Are these mutations very common in many cancers? I feel like if that were the case they would has also been accurate in pancancer-only.

Is the takeaway that cancer types have very distinct mutational signatures and so removing the cancer type of interest performs very poorly. But then is there an advantage to having some pancancer information?

pancancer_evaluation/utilities/classify_utilities.py
```diff
@@ -66,14 +68,24 @@ def run_cv_cancer_type(data_model, gene, cancer_type, sample_info, num_folds,
...
+                   seed=data_model.seed)
         except ValueError:
             raise NoTestSamplesError(
                 'No test samples found for cancer type: {}, '
                 'gene: {}\n'.format(cancer_type, gene)
             )
 
+        if X_train_raw_df.shape[0] == 0:
+            # this might happen in pancancer only case
+            raise NoTrainSamplesError(
```
ajlee21, line 84: If this error is only raised for cancers with 0 samples, maybe update the text to say 0 samples instead of insufficient samples. I at least thought that you had some threshold you were using to determine suffciency

tests/test_cross_validation.py
```diff
@@ -55,8 +58,30 @@ def test_cv_pancancer(expression_data, cancer_type):
     test_cancer_types = get_cancer_types(
             sample_info_df, test_df_pancancer.index)
 
+    assert len(train_cancer_types) > len(set([cancer_type]))
     assert set([cancer_type]).issubset(train_cancer_types)
+    assert test_cancer_types == set([cancer_type])
+
+
+@pytest.mark.parametrize("cancer_type", ['BRCA', 'COAD', 'GBM'])
```
ajlee21, line 66: Nice tests. I need to learn how to make modular test cases

ajlee21 (approved):
Code all looks good to me. Just some conceptual questions

05_plot_cancer_type_results.ipynb
```diff
@@ -559,7 +560,7 @@
     "sig_genes = list(set([id_str.split('_')[0] for id_str in sig_ids]))\n",
     "\n",
     "# then get overlap with genes that are significant in stratified analysis\n",
-    "vogelstein_results_df = pd.read_csv(os.path.join(cfg.results_dir, 'vogelstein_stratified_pvals.tsv'),\n",
+    "vogelstein_results_df = pd.read_csv(os.path.join(results_dir, 'vogelstein_stratified_pvals.tsv'),\n",
```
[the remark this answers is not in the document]
jjc2718: Great questions! These experiments are getting progressively more complicated, so it's super helpful that you're checking that all of this makes sense. See answers to your questions below:

> Your volcano plots are comparing genes that were predictive of mutation status in a single cancer A using 1) training set of a single cancer A vs shuffled dataset and 2) training set of pan-cancer including cancer A vs shuffled dataset. So these plots are showing there are are indeed a subset of genes (in orange) that seem to be relevant for predicting mutation status in cancer A

This is the right idea, but I'd change the wording a bit: "comparing genes that were predictive of mutation status" -> "comparing genes whose mutation status could be predicted from gene expression". So the genes in the volcano plot are the *labels* in our models (not to be confused with the gene features selected as being predictive by the model; we haven't analyzed these in detail yet, although that will likely be my next PR).

Once you make that change, what you said is exactly right.

> These venn diagrams are showing the predictive genes (to predict mutation status of a single cancer) found using single cancer training set (I think this is what you're calling your "Cancer type only") vs using stratified cancer training set. You also have a venn diagram showing the predictive genes found using pan-cancer vs using stratified cancer.

Again, small change: "predictive genes" -> "genes whose mutation status can be accurately predicted"

> I think your label for the venn diagram using pan-cancer data may need to be updated if my understanding is correct.

Good catch! I'll fix this.

> So is the takeaway here that there are some predictive genes that are predictive no matter what training dataset you start with (single, stratified, pan). But there are some genes that are found to be predictive using single and pan cancer data for training but not in stratified. Is this what you would've expected? I'm guessing that the overlap genes are those that have a robust contribution to any cancer mutation status. But the non-overlap ones for single cancer are those that have a specific contribution. I'm not sure how to explain the pan-cancer non-overlap ones.

Yes, this is exactly what I would have expected (and I agree with your summary of the takeaway):
* For any gene that can be predicted well in the stratified data set, we would expect that there are one or more individual cancer types where it can also be predicted well. This is what we see, there are almost no "stratified only" genes (except for JAK2 in the single cancer case, but this is probably just due to a lack of statistical power/limited data in individual cancers, I'm not too worried about it. If we saw a lot of these I'd be more concerned/confused).
* It's pretty widely reported that certain mutations can have strong driver effects in certain tissues/cancer types, and weaker or no effects in others. So, when we split by cancer type, we would expect that some genes show up as significant that weren't significant in the stratified dataset (i.e. the signal was "washed out" when we mix all cancer types together, but when we split cancer types out individually we see stronger effects).
* The pan-cancer non-overlap genes are less intuitive, but I think it's likely that these are just being driven by the strong signal in one or two cancers. I'm not surprised that we see results here, although it may be interesting to look further into these results (i.e. what genes/other covariate information do these rely on? if we remove certain cancer types, do these signals disappear?)

> What is the difference between pan-cancer vs stratified? Is it that they both have mixed cancer types but stratified controls for having uniform proportions of cancer and pan-cancer doesn't?

Yep, exactly. In the stratified experiments, the test set and training set both have uniform proportions of each cancer type. In the "pan-cancer" experiments, the test set is composed only of a single cancer type, and the training set is a mix of that cancer type and all the rest of the cancer types in TCGA (subject to our filters for label imbalance, etc).

So to put it slightly differently, the goal of the stratified experiments is to identify **genes** whose mutation status we can predict from gene expression, across a mixture cancer types. The goal of the "pan-cancer" experiments, on the other hand, is to identify individual **cancer types** where this prediction works well.

06_plot_remove_cancer_results.ipynb
jjc2718: Again, great questions! Thanks for diving deep into this, it really helps.

> So it looks like you have some genes that very accurately predict mutation status using pancancer but not using pancancer-only. And these genes are not signficantly more accurate compared to your shuffled data. What would you say about these genes? Are these mutations very common in many cancers? I feel like if that were the case they would has also been accurate in pancancer-only.

I don't know if there's anything especially significant about the genes that show a strong difference in performance without the data from the relevant cancer type. I only ran one replicate of 4-fold CV here (for the experiments in the other notebook I ran 2x4-fold CV), so I suspect that with an additional replicate we'd see a lot more significant genes on the left of these plots.

Mostly I was interested in the general trend (to the left or roughly equal), but that appears to be behaving the way we expected (shifted to the left in these volcano plots) so I think I'll leave it at that for now and move onto other things.

> Is the takeaway that cancer types have very distinct mutational signatures and so removing the cancer type of interest performs very poorly. But then is there an advantage to having some pancancer information?

Yeah, basically - this experiment was mainly a sanity check to see that in most cases, removing the cancer type we're testing on from the training set hurts performance. The volcano plots show pretty clearly that this is generally the case, so that makes us more confident that things are running correctly.

To answer your second question, I would say yes, in general there is usually still an advantage to having pan-cancer information, even if it's not the right cancer type. See, for instance, this example from the notebook:

image

If you look at the `pancancer_only` box plot, you can see that the model clearly outperforms the shuffled baseline, even without data from the relevant cancer type. So there does seem to be some signal in data from other cancer types, at least in the case of BRAF (and for most of the other genes I've checked by hand).

The best way to use this in a pan-cancer model, however, is unclear. See this plot from `05_plot_cancer_type_results.ipynb`:

image

Gene/cancer type combinations on the right have better performance when pan-cancer data is used. There are very few examples that significantly outperform the single-cancer models, although there are a small handful.

Taken together, these results suggest that there is useful signal in the data for unrelated cancer types, but maybe we're not using it in the right way with our simple linear model. It's possible that better models or different approaches to incorporating data from unrelated cancers would help to shift more points on this plot to the right. This is something Casey and I have been thinking about and we have a few ideas, so stay tuned.

PR #31, opened 2020-10-15
Model coefficient stability analysis

jjc2718:
The goal of this analysis was to use the model coefficients of our mutation prediction classifiers to evaluate similarity between models. Since we're using elastic net logistic regression (which zeroes out coefficients for most genes), we can compare the nonzero coefficients between models, and if they are similar we say the models are similar.

The idea was to eventually use this to define similarities for the same gene across different cancer types (e.g. if we noticed that our KRAS mutation predictor selects similar genes in thyroid cancer and colon cancer, we would hypothesize that KRAS mutations have similar effects on gene expression in those cancer types, which could be interesting biologically).

Unfortunately, this doesn't work as well as we thought it would - even for models on different cross-validation folds of the same gene and cancer type, we see considerable variation in the nonzero coefficients. This is probably due to the large amount of multicollinearity in gene expression data: in many cases there are multiple predictors/genes in the dataset conveying essentially the same information, so the model can pick one or a few of them essentially arbitrarily.

This is a fairly well-documented characteristic of feature selection in linear models on datasets with collinear features, so it isn't too surprising.


pancancer_evaluation/utilities/analysis_utilities.py
```diff
@@ -72,6 +73,42 @@ def generate_nonzero_coefficients(results_dir):
...
+    """
+    coefs = {}
+    all_features = None
+    gene_dir = os.path.join(results_dir, gene_name)
+    if not os.path.isdir(gene_dir): raise StopIteration
+    for coefs_file in os.listdir(gene_dir):
+        if coefs_file[0] == '.': continue
+        if 'signal' not in coefs_file: continue
+        if 'coefficients' not in coefs_file: continue
+        cancer_type = coefs_file.split('_')[1]
```
ben-heil, line 102: Magic numbers are bad, but this one is probably harder to fix than it's worth. Maybe making a note in the scripts that generate coefficient files that the cancer type must be the second field would help keep this from breaking.

07_coefficient_analysis.ipynb
```diff
@@ -821,6 +822,460 @@
...
+    "\n",
+    "In the single-cancer case, model overlaps are much lower in general, and there is not nearly as much separation between well-performing and poorly performing models. This is a bit surprising since we're only considering models that predict mutations in the same gene, so we'd expect them to use some of the same predictors, but this doesn't really seem to be the case often. This may be due to the high level of correlation/collinearity in gene expression data (there are multiple choices of predictors that can predict the same labels, since many of the predictors convey redundant information).\n",
+    "\n",
+    "Maybe looking at this from a pathway or network perspective would help to elucidate this; that is, even if the model selects different genes between folds, we'd expect them to be in the same pathway, or to be close together in a gene interaction network."
```
ben-heil, line 1277: Aggregating genes into networks/pathways is a good idea. You could also try some form of dropout or masking on the inputs to see which genes have similar weights vs swing depending on how many of their colinear partner genes drop out

ben-heil (approved):
Looks good! One potential way you could cut down on colinearity is by using a subset of genes like the LINCS1000 landmark genes (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5990023/). I think your plan to aggregate at the network/pathway level is probably better though

pancancer_evaluation/utilities/analysis_utilities.py
jjc2718: Will do! I agree the magic number isn't ideal, but as long as I use the same convention everywhere (identifier = {gene}_{cancer_type}), I don't think it will cause issues.

I'll definitely make a note in the scripts that generate these files, though.

07_coefficient_analysis.ipynb
jjc2718: This is a good idea! I'm not sure I'll get to this in the near future, but I'll create an issue to remind myself, since I think this would be a neat analysis once we have some results we're more confident in.

pancancer_evaluation/utilities/analysis_utilities.py
```diff
@@ -263,3 +311,58 @@ def get_mad_proportion(coefs, mad_genes):
...
+        for f1, f2 in it.product(range(num_folds), repeat=2):
+            try:
+                nz_coefs_1 = list(zip(*coefs_list_1[f1]))[0]
+                nz_coefs_2 = list(zip(*coefs_list_2[f2]))[0]
+                fold_jaccards.append(compute_jaccard(nz_coefs_1, nz_coefs_2)[0])
+            except IndexError:
+                # this can occur if all coefficients for a given fold were zero
+                # (i.e. model predicts the mean/fits only an intercept)
+                # if so, we call it a jaccard index of 0
+                fold_jaccards.append(0.0)
```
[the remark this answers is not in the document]
jjc2718: Looking at these examples (there's only 8 genes where this happens out of ~120), all of these gene/cancer type combinations seem to be instances where the model doesn't outperform the negative control. In other words, for these genes and cancer types there isn't any predictive signal at all in the gene expression data, so I'm not surprised that no genes are selected.

I'm actually a bit surprised it doesn't happen more often in these poorly performing cases. My guess is there are some cases where our non-gene expression covariates (sample mutation burden or cancer type) provide a small amount of predictive signal, but not enough to outperform the baseline.

pancancer_evaluation/utilities/analysis_utilities.py
```diff
@@ -263,3 +311,58 @@ def get_mad_proportion(coefs, mad_genes):
...
+        try:
+            id1_sig = pancancer_comparison_df[
+                pancancer_comparison_df.identifier == id1
+            ].reject_null.values[0]
+        except IndexError:
+            # if identifier isn't in statistical testing results for some reason,
+            # assume it isn't significant
+            id1_sig = False
```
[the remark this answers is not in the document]
jjc2718: Thanks for pointing this out. After looking into this I do think this is a (somewhat subtle) bug. Basically, the issue is that for genes with very few positive (mutated) samples, sometimes, but not always, the randomly shuffled cross-validation splits will result in a split with samples from only one class.

Right now I'm skipping them in my experiment code, but because the cross-validation splits can vary between seeds, sometimes the model fit works for one seed (present in the analysis) but not another (not present in the pancancer df).

I'm not sure what the best solution to this is. I was aware of this previously and had issue #4 to stratify label proportions in cross-validation, so maybe revisiting that is the best solution. I'll create a future PR if that's what I decide to do, and for now I'll just keep in mind that I could be missing some genes from this analysis.

PR #34, opened 2020-10-30
Optimizations for cross-cancer experiment scripts

jjc2718:
No new results in this PR, just changes to how I'm running some of the experiments. I've made sure that this PR doesn't change the results of the experiments at all, but things are substantially faster now.

Essentially, the cross-cancer experiments I was running before were filling in entries in a matrix by training a new model for each matrix entry. I realized this didn't make sense - since each row of the matrix trains a model on the same gene/cancer type, we can just train the model once per row, then evaluate it lots of times to fill in that row (prediction/model evaluation is much faster than training).

This is a bit more complicated for the pan-cancer experiments, since we actually need to train different models for some matrix columns depending on which cancer type is held out, but it's possible to cache pre-trained models to help in this case as well.

There are more details about the actual optimizations in the code comments, if you're interested.


pancancer_evaluation/data_models/tcga_data_model.py
```diff
@@ -232,6 +232,47 @@ def process_data_for_gene_id(self,
...
+                                    test_cancer_type=None,
+                                    shuffle_labels=False):
+        """
+        Prepare to train model on a given gene. Preparation of test data must
+        be done later (e.g. by calling another process_* function).
+
+        Arguments
+        ---------
+        train_identifier (str): gene combination to train on
+        train_classification (str): 'oncogene' or 'TSG' for the training gene
```
ajlee21, line 248: I'm not sure what this training gene is? Is this a "negative control gene"?

pancancer_evaluation/data_models/tcga_data_model.py
```diff
@@ -232,6 +232,47 @@ def process_data_for_gene_id(self,
...
                 self.y_test_df.status.values)
 
+    def process_train_data_for_gene(self,
+                                    train_gene,
+                                    train_classification,
+                                    output_dir,
+                                    test_cancer_type=None,
+                                    shuffle_labels=False):
+        """
+        Prepare to train model on a given gene. Preparation of test data must
```
ajlee21, line 242: Don't know if you want to add a bit more in the description about how the data is being prepared? Looks like you're splitting the data as well as making sure the training set includes gene expression data from only one gene and cancer type or one gene and any cancer type

ajlee21 (approved):
Very neat code. Glad you were able to make this run faster!

03_cross_cancer_classification/run_cross_cancer_classification.py
```diff
@@ -73,35 +76,43 @@ def process_args():
...
+        print('shuffle_labels: {}'.format(shuffle_labels))
 
-        progress.set_description('train: {}, test: {}'.format(
-            train_identifier, test_identifier))
+        outer_progress = tqdm(identifiers,
+                              total=len(identifiers),
+                              ncols=100,
+                              file=sys.stdout)
+
+        for train_identifier in outer_progress:
```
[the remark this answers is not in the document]
jjc2718: Yep, exactly - a few lines below that I call `train_cross_cancer` to train a model for each row, which gets saved in the `model_results` variable.

Then a few lines later, I loop over all the test identifiers (columns) and call `evaluate_cross_cancer` once for each of them, taking `model_results` as a parameter. This just evaluates the model (i.e. makes predictions for all the test labels and calculates the ROC and precision/recall).

pancancer_evaluation/utilities/classify_utilities.py
```diff
@@ -27,20 +27,18 @@
     NoTestSamplesError
 )
 
-def classify_cross_cancer(data_model, train_identifier, test_identifier,
-                          shuffle_labels=False, train_pancancer=False):
+def train_cross_cancer(data_model,
+                       train_gene_or_identifier,
```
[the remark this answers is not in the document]
jjc2718: This function is called in both the `run_cross_cancer_classification.py` and `run_pan_cross_cancer_classification.py` scripts. It's a bit confusing since each of those contexts is slightly different (for the cross-cancer experiments we're training on a single gene/cancer type combination, which I'm calling an "identifier", and for the "pan-cross cancer" experiments we're training on all the data for a single gene).

The main difference in this PR is that I separated this function from the `evaluate_cross_cancer` function (before they were a single `classify_cross_cancer` function). This means I'm able to pull `train_cross_cancer` out of the inner loop in the run* scripts.

Does that make sense, or am I misunderstanding what you're asking?

pancancer_evaluation/data_models/tcga_data_model.py
jjc2718: Ah, sorry, this was a copy/paste typo (should say `train_gene` rather than `train_identifier`).

This function separates the data into train/test datasets for the "pan-cross cancer" experiments. So the idea is to train the model on all cancer types but one, and test the model on the remaining cancer type. `train_gene` here just tells this data processing function which gene to get the labels for (and consequently which cancer types to use).

jjc2718: Oh, wait - are you asking what TSG means here? It's an abbreviation for "tumor suppressor gene", not a particular gene. I'll write that out here since it definitely won't be clear to non-cancer people.

pancancer_evaluation/data_models/tcga_data_model.py
jjc2718: Yep, good call - this function is doing a few different things. I'll add some clarification.

pancancer_evaluation/data_models/tcga_data_model.py
ajlee21: you answered correctly, I thought "TSG" was an abbv for "training gene" thus the phrasing of my question.

PR #36, opened 2020-11-17
"Flip labels" positive control experiments

jjc2718:
**Note:** `04_coefficient_analysis.ipynb` doesn't need to be reviewed, I'm just adding back changes that I (somehow) accidentally deleted in a previous PR.

In general, the goal of this PR is to "flip" or hold out a subset of positively labeled samples in a given gene and cancer type, and see if a model trained on the rest of the samples can differentiate the "flipped" samples (false negatives) from the true negatives.

For an idea of how I'm training/testing, I made this extremely high-tech and polished graphic:

cropped_training_image

(shaded samples have positive labels, the others have negative labels). What this is showing is that I'm removing the positively labeled samples from the test set, since including them as positives would inflate performance (training on test data) and including them as negatives (what we did at first) would artificially deflate performance.

Results are in `03_cross_cancer_classification/plot_flip_labels_results.ipynb`.


03_cross_cancer_classification/plot_flip_labels_results.ipynb
```diff
@@ -0,0 +1,546 @@
...
+    "percents_flip = flip_norm_df.percent_flip.unique()\n",
+    "for ix, percent_flip in enumerate(percents_flip):\n",
+    "    ax = axarr[ix // 2, ix % 2] \n",
+    "    heatmap_df = (\n",
+    "        flip_norm_df.query('percent_flip == @percent_flip')\n",
+    "                    .pivot(index='gene', columns='cancer_type', values='aupr')\n",
+    "    )\n",
+    "    sns.heatmap(heatmap_df, vmin=0.0, vmax=1.0, ax=ax)\n",
+    "    ax.set_title('Percent of labels flipped: {}'.format(percent_flip))\n",
+    "plt.tight_layout()"
```
ben-heil, line 368: I think there's supposed to be a heatmap here, but it doesn't show up when rendering the notebook

03_cross_cancer_classification/plot_flip_labels_results.ipynb
```diff
@@ -0,0 +1,546 @@
...
+    "for ix, plot_gene in enumerate(plot_genes):\n",
+    "    ax = axarr[ix // 3, ix % 3] \n",
+    "    plot_df = flip_norm_df[flip_norm_df.gene == plot_gene]\n",
+    "    sns.pointplot(data=plot_df, x='percent_flip', y='aupr', hue='identifier', ax=ax)\n",
+    "    ax.set_xlabel('Percent of 1 labels flipped to 0')\n",
+    "    ax.set_ylabel('Change in AUPR vs. negative control')\n",
+    "    ax.set_title(plot_gene)\n",
+    "    ax.get_legend().remove()\n",
+    "    ax.axhline(y=0, color='black', linestyle='--')\n",
+    "plt.tight_layout()\n",
```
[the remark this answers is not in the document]
ben-heil: Also the results seem to be bimodal for a couple genes (PTEN and KRAS). Is there a reason for different cancer types to be more or less able to differentiate between mutated and not? Are PTEN and KRAS passenger mutations instead of drivers in the cancer types that perform less well?

03_cross_cancer_classification/run_cross_cancer_flip_labels.py
```diff
@@ -0,0 +1,239 @@
...
+    identifiers = ['_'.join(t) for t in it.product(sampled_genes,
+                                                   tcga_cancer_types)]
+
+    # create output directory
+    output_dir = Path(args.results_dir, 'cross_cancer_flip_labels').resolve()
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    for shuffle_labels in (False, True):
+
+        print('shuffle_labels: {}'.format(shuffle_labels))
```
ben-heil, line 103: I'm not sure if this is a debug print statement or part of the desired output. If it's for debugging you can probably remove it

ben-heil (changes requested):
Looks good, but there are a couple issues in the viz notebook that you'll want to fix before merging

03_cross_cancer_classification/run_cross_cancer_flip_labels.py
jjc2718: It's part of the desired output - when my experiments are running it prints above the progress bars from `tqdm` so I can tell whether I'm using shuffled or original labels.

I guess I could make it into another (nested) progress bar, but I already have two here, and it seems to me like having a progress bar for only two steps wouldn't provide much information.

03_cross_cancer_classification/plot_flip_labels_results.ipynb
jjc2718: Nice catch! I'll fix this.

03_cross_cancer_classification/plot_flip_labels_results.ipynb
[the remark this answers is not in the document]
jjc2718: > Is there supposed to be a plot in the bottom right subplot?

Nope, there's just an odd number of subplots so I have an empty grid section. I'll get rid of the unused axes so this is clearer.

> Also the results seem to be bimodal for a couple genes (PTEN and KRAS). Is there a reason for different cancer types to be more or less able to differentiate between mutated and not? Are PTEN and KRAS passenger mutations instead of drivers in the cancer types that perform less well?

This is a good question that I've been thinking about myself. My guess isn't necessarily that they're not drivers in these cancer types, but rather that they're very rarely mutated in the dataset that I'm using (very imbalanced labels) in these cancer types, so classification doesn't work well. I've been thinking about doing an analysis that looks at label imbalance vs. performance in more detail, so stay tuned on that.

Your thought (they may just not be drivers in these cancer types) is definitely a possibility as well; some drivers are definitely thought to be cancer type-specific. Both PTEN and KRAS are thought to have fairly ubiquitous driver effects, though - looking at Figure 2 from [this paper from TCGA](https://doi.org/10.1016/j.cell.2018.02.060), it looks like they're both predicted to be drivers in multiple cancer types, including some of the cancers that perform poorly in my analyses. I'm pretty sure their driver prediction analysis primarily just uses mutation frequency in DNA sequencing data, so they may be better powered to detect driver effects in these cancer types than my analysis which requires paired DNA/RNA (although I'm not super familiar with their methods so I'm guessing a bit at this point). Definitely interesting stuff to think about!

03_cross_cancer_classification/run_cross_cancer_flip_labels.py
ben-heil: That makes sense! Was just making sure it was intended

03_cross_cancer_classification/plot_flip_labels_results.ipynb
```diff
@@ -0,0 +1,559 @@
...
...kkxVqB0KwtD7xQQFDTh7llgiAIgiAInSIGMCorK9mwYQN//vOf8XhCBc4+/PBDLr300iFrnCAIgtCV6Jt7p+sGj7yyl/qWUEBnTlYil62Y3OO281PmhF/vri0YkvaNZPHRFu64dj6xUaH6F4WHG3nklb3o+ogfaSoIgiAIwhgRsQbGNddcwxVXXMGxY8fw+/2YTCbefvttNmzYwKJFi4a0kSN9nLUgCEJ/ne5Ya9E39+6FLSW8tq0MgJQ4G3d9fRGOCLUcDMPg19t/R427DpNs4v6Vv8CiiOKVlXVO7n96N05PAIAVc9K46ZIZSJI0zC0ThMEhamAIgiCcOSLOQiJJEuvWrQPgvPPO46yzzuLll1/GYrEMWeMEQRCErkTfHNmu4rpw8MJskvnelbkRgxcQ+l3OT87lzbL3CegB9jYcYMFxWRljVXpyFD+6ei6/fWY3Xr/G1vwq7BaVdedNFUEMQRAEQRCGVcQhJKraGduIi4vjvvvuExfIgiAIw+xU+2Zd17n77rtZt24dN9xwA2VlZV3Wv/XWW6xdu5avfOUrPPfccwPe7sFW1eDif1/bF35/0yUzmZAS1efnug4jyR+Utp2JJqfF8IOvzMGkhi4T3v68gle3HRneRgkRabrOR3uOcc/Gndzx4Mfcs3EnH+05Job/CIIgCKNOrxkYHaKiosRTF0EQhBHgVPvmd999F7/fz7PPPkteXh733XcfDz30EACapvH73/+ef/3rX9jtdtasWcP5559PQkLCoJzDQPP4gvz1hQK8/tCsGRedNZHFs1L79dn0qDSSbYnUeRoobDiAX/NjFsNIAJieEc93Ls/hry8UoOkGL310GJtF5cJFE4e7acJxNF3n4Zf2srO4LryssdXHocoW8ksauPXy2SiymHROEARBGB0iBjB27drFihUrAGhubg6/Bti6devgt0wQBEHo5lT75p07d7Jy5UoA5s2bR2FhYXidoii88cYbqKpKQ0MDAA6Ho8+2KIpEXJy9z+0Gk2EYPPLUTqoaQrU4cqYk8s3Lc1GU/t+wLZm4gFeL38Gv+SnzHuGs9HmD1Nozz+pFGciqwh+f3Y1hwDPvHqS22cPROhcNzR4S42xcsGgi5yyciCKLBx3D4b3Py7sEL463s7iOvNImzhdBJ0EQBGGUiBjAOP7i9mTpus769espKirCbDazYcMGMjMzw+tfe+
```
ben-heil, line 527: Idk how to do it but you could probably put the color key in its own panel in the bottom right

ben-heil (approved):
Looks great!

03_cross_cancer_classification/plot_flip_labels_results.ipynb
jjc2718: Yeah, I think if this were a figure for a publication or something like that I'd try to put a color key there, but these experiments are mostly just a sanity check so I don't want to spend too much time on it now (and my gut feeling is that it would take a decent amount of tweaking/trying things to get it to work right).

If I end up using this code (or similar code) in future analyses, I'll go back and fix the key/legend.

PR #37, opened 2020-11-30
New cross-validation scheme + more positive control experiments

jjc2718:
In #36, I did some experiments where I flipped positive labels to negative, then tried to predict them in the test set. I realized that a possible reason this wasn't as bad for performance as we expected was that the true negatives are the same between train and test sets.

So, in these new experiments I'm holding out both positive and negative labels, then testing on the holdout data:

holdout_labels_schematic

This experimental setup should be more reflective (I think) of the actual effect that limiting training data is having on our models. Unfortunately the results aren't much clearer than before, but we're planning to move on after this to something different, since we're a bit in the weeds here and we're not really seeing the behavior we expect.


ajlee21 (commented):
I think I'm a little lost here. So say you have samples A,B,C,D from some cancer/gene pair. You have the associated labels [+,+,-,-]. In your previous experiment you flipped some percentage of labels from + to -. So you have [+,-,-,-]. What is it you're trying to predict with your training model? You're trying to predict the label of mutation +/- based on gene expression data? You were expecting to see a reduction in performance given you are training are more inaccurate labels as you flip more, but you didn't see that.

What is the goal or takeaway of this experiment?

Some questions and comments:

* Looks like you have some comments leftover from the previous analysis -- talking about flipped labels.

* For the heatmap plots, its a little hard to visually compare across the different panels. I wonder if it would make more sense to plot the gradient instead so you can focus on the change per cancer/gene. But it looks like you're better able to see these trends on your line plots. Since this is more exploratory I don't think you need to change this, but just a thought for future experiments. 

* (minor) What do the grey boxes mean in your heatmap?

* So you'd expect to see a reduction in your performance (i.e. your ability to detect if a gene has a mutation in a cancer type given gene expression data for that cancer and gene), but you don't, even after removing 95% of samples from training? And this is especially surprising given that there is a correlation between the number of positive samples and the change in performance after removing 95% of the data (i.e. the greatest difference in performance is associated with those gene/cancers with the most positive cases)

These results are definitely interesting....(i apologize in advance if you've already discussed this, i realize you're trying to move forward so please feel free to ignore or move to issues): Just trying to take a step back to figure out what is happening here. Can you tell me what the training set is again? I wonder if its possible for you to simplify this experiment even more so that you're not dealing with lots of features/genes in your training set (i.e. don't use pancancer)? How much data is left for training after 95% is removed? I assume you've already checked that your labels are balanced after removing data.

jjc2718:
Thanks for the feedback! We're pretty far in the weeds at this point, so it's not surprising that you're a bit confused. I'll do my best to answer some of your questions.

> I think I'm a little lost here. So say you have samples A,B,C,D from some cancer/gene pair. You have the associated labels [+,+,-,-]. In your previous experiment you flipped some percentage of labels from + to -. So you have [+,-,-,-].
What is it you're trying to predict with your training model? You're trying to predict the label of mutation +/- based on gene expression data? You were expecting to see a reduction in performance given you are training are more inaccurate labels as you flip more, but you didn't see that.

A few points that I think may help clarify:

- At this point, we're training and testing on a single gene in a single cancer type. These are supposed to be "positive control" experiments since the cross-cancer ones didn't really work (i.e. we couldn't really predict the same gene across different cancer types well, with some exceptions). The rationale was that if we can't predict held-out labels within a cancer type, we're wasting our time looking for relationships across cancer types.
- Yep, we're still trying to predict mutation labels based on gene expression data. We picked a few genes to focus on here that we've seen good performance on in the past.
- Using your example, say the true labels were [+, +, -, -]. In our old experiments, we'd train on [+, -, -, -] and test on [+, +, -, -] under the assumption that our model should be able to recognize the false negative (sample 2) as actually being positive, given enough true positives (sample 1) and true negatives (samples 3 and 4). For the most part this was successful, but we were a bit concerned that performance didn't seem to degrade much when we flipped almost all of the positives to negatives in the training set.
- I hypothesized that the problem was the true negatives: no matter how many positives we flipped we were still telling our model what a sample without a mutation looks like. So I switched (in this PR) to doing a more traditional train/test set split. For your example, that would look something like training on samples 1 and 3: [+, -], then testing on samples 2 and 4: [+, -] (with varying train/test set sizes).
- And, as you can see this still didn't seem to degrade performance all the way, although it helped somewhat. We tried a few other things to break the model: shuffling the training data worked (duh) but most other things didn't.

> What is the goal or takeaway of this experiment?

Goal: better understand model failure cases, for relatively easy examples (i.e. genes and cancer types where we know there is predictive signal

Takeaway: some genes (TP53 in particular) are *very* resilient to reductions in sample size, and can generalize from only tens of samples. Others (KRAS, EGFR) do degrade noticeably but not all the way to 0. This leads us to think our inability to detect cross-cancer signal is an actual limitation of our data or modeling approach, not some kind of artifact or bug.

> Some questions and comments:
> 
> * Looks like you have some comments leftover from the previous analysis -- talking about flipped labels.

Updated these!

> * For the heatmap plots, its a little hard to visually compare across the different panels. I wonder if it would make more sense to plot the gradient instead so you can focus on the change per cancer/gene. But it looks like you're better able to see these trends on your line plots. Since this is more exploratory I don't think you need to change this, but just a thought for future experiments.

Good idea! I think since we're moving away from these experiments I'll leave it for later, but I'll definitely keep it in mind for the future.

> * (minor) What do the grey boxes mean in your heatmap?

These are just gene/cancer type combinations where the labels are too imbalanced for our filters (we require at least 10 samples to be mutated, and these must make up at least 5% of the dataset). Not every driver gene is frequently mutated in every cancer (although some, like TP53, are mutated in many cancers).

> * So you'd expect to see a reduction in your performance (i.e. your ability to detect if a gene has a mutation in a cancer type given gene expression data for that cancer and gene), but you don't, even after removing 95% of samples from training?

Yep, exactly!

> * And this is especially surprising given that there is a correlation between the number of positive samples and the change in performance after removing 95% of the data (i.e. the greatest difference in performance is associated with those gene/cancers with the most positive cases)

My argument was actually the opposite, that this correlation explains some or most of the difference in performance, which would make it less surprising (for example, there are a lot of samples that have TP53 mutations, so it makes sense that it's more resilient to removing samples since there's more signal to start with). Removing 95% of samples is a lot, though, so it's still not exactly what we expected (the TP53 line plot is essentially flat, we'd expect to see at least some performance decrease).

I think I answered most of your other questions above, except these:

> How much data is left for training after 95% is removed?

Good question - I haven't done this analysis yet, but maybe that would help. I think the correlation plot is probably a decent proxy for the actual amount of data left, though.

> I assume you've already checked that your labels are balanced after removing data.

I'm removing the given proportion of samples from positive and negative sets separately, so all of the train/test sets will be as balanced as the input data from TCGA was for that gene/cancer type combination. Some of these combinations are more imbalanced than others, though.

I do think label imbalance probably has a substantial effect on the variation between results we see here, between genes and between cancer types. I've been thinking about analyses to look further into this, but I haven't had any great ideas yet. I'll create an issue to remind myself to come back to this in the future.

ajlee21 (approved):
Thank you for all the comments. I understand what is happening now and I agree with your conclusions. Very strange behavior!

Just some thoughts for the future that we discussed offline (so you have a record):
* Maybe try looking into a different eval metric to better handle class imbalance
* I wonder if some feature selection step before training the model would help in case signals are getting washed out except for those very salient ones like TP53
* Maybe its how you're subsetting labels?
* Of course double checking that there isn't a bug somewhere in your approach if you haven't already, which I suspect you thoroughly have.

jjc2718:
A few other things I discussed offline with @ajlee21 :
* Could plot a PCA/UMAP etc. of training gene expression data in the case where 95% of data is held out, and check if data is easily separable by label (see issue #39)
* Would 100% of flipped labels have the same effect as 100% of shuffled data? (I think no because some labels will be correct by chance with shuffled data, but haven't looked into this in depth)

I'm going to merge this PR in the interest of moving on to other things, but I think there's a lot of loose ends here that we could explore in the future.

PR #40, opened 2020-12-15
Positive control/cancer type holdout experiments follow-up

### LINKED ISSUE #39: Dimension reduction in 95% of held-out data case
See #37 (most recent comment, specifically): could do a PCA or UMAP plot to see if data is easily separable, could take subsets, etc.

jjc2718:
This PR implements a few of the ideas that were brought up in my braintrust presentation last week.

Specific changes in this PR:

* Plot label count rather than holdout percent on x-axis of line plots (suggested by @ben-heil)
* PCA/UMAP plots of train/test sets in holdout experiments, to see if well-performing cancer types are linearly separable (suggested by @ajlee21 and @nrosed; closes #39).
* Added cleaned-up version of line plots with more legible cancer type labels (don't remember who suggested this, probably multiple people since the old version was impossible to read).

I ran the PCA/UMAP script a few times to compare different genes/cancer types, and put the results in [these Google slides](https://docs.google.com/presentation/d/1Q1LkqVqnr82twaFJ_aYYjmjno3cqA7j17VwyUZzECPs/edit?usp=sharing) (should be globally viewable, I think).

These results give me slightly more confidence that what we're seeing in these experiments is an actual property of the data for certain genes/cancer types, and not just a bug or a limitation of our performance metrics.


ajlee21 (commented):
Some comments for `03_cross_cancer_classification/plot_holdout_single_cancer_results.ipynb`

* Really nice description in the header. Its very clear what this notebook is going to do
* Just to clarify, when you say "percent_flipped" = percent held out of training? I assume this is just borrowing old terminology so I just wanted to confirm.
* By eye it looks like the trends are similar between the old line plots (proportion of held out) vs new line plots (raw number held out). Can you remind me the logic for this additional plot? I assume that different cancer types will have different numbers of total samples, so I feel like proportion makes more sense to look at in this case. But I'm sure it was mentioned during your BT and I have forgotten. So maybe 95% of the data in cancer A is 400 but cancer B is 100. What does looking at the raw data tell you? I feel like you should just see curves drop off as you move from left to right, but not sure what else...
*

03_cross_cancer_classification/plot_holdout_pca.ipynb
```diff
@@ -0,0 +1,385 @@
...
...+vUb6I47Wqh27Tru/x58sK8+/XSrEhMvljqOtm3ba+/eb5SQkKDIyCjdddfd2rXrX8rKynIP8S1Mr1736513/qaEhAT9+usFvffeiiL39/evptTUVPfrdu06KCHhvP7xj/Xq0aNXnn3Xrv27Tp8+pZSUZC1e/Ka6d++hatWqqXfvPnr//ZX6+ef/k8vl0t//vlr//d8jlZ7umWHh5ZVy1xq5gprJsDnkCmqmlLvWVNqxrXKNlUaDBg21bdsWZWVlas+er/XDD/92b/P393dfP6mpKbLbHapRo7qcTqc2b/6nDhz4rkSrE/v7V1NGxpVyLbwCAAC8Dz0IAQBAoRo3Njw252B+o0aNkN1uk81mk59fNd19d2vNmjW7yJ955pmJGjKkvxYv/qv279+n7t17XrPP3Xf/p0JDw7Rx43rdemvhyaCC1K5dW/Xr19f11/9BNptN9es3UFhYWJGLk+QYPXqMUlNTNXRof9WoEaAePXrlWXgkv+7de2nBgnk6e/aMpk2bqWrVqqlLl3v1+eefqW3bdnn2vfXW2zR5crTOnz+nTp26KDo6exXlXr3u16VLSYqOflq//fabGjdurAULFuUZwmolrsAIj845mJ8Vr7HSmDRpihYtWqD169fqrrta54mlXbsOWrNmtQYOfFirV69T167dNGzYQNntDt10003q3buP4uOLXwQmKqqZIiIi1aNHF7377vtq0KChx84HAABYh80wDCYcAQCgijp06EfdcIP15p5Dybzzzt90/vw5TZnyP+73xo0bo65d79WAAYNNjAy+6syZk7rllpvNDgMAAFQwhhgDAABYzMWLF3Xo0A/asOGjPPMJAgAAAJ5AghAAAMBiDhz4VuPHP65u3brr5ptvMTscAAAAVHEMMQYAoApjiDGAisQQYwAAqiZ6EAIAAAAAAAA+jAQhAABVHIMFAFQE6hIAAKouEoQAAFRhDodDTmeW2WEAqAKcziz5+TnMDgMAAHgACUIAAKqwWrXCdenSRRmGy+xQAHgxw3ApKemiwsPDzQ4FAAB4AIuUAABQhblcLv38889KTU0Vf/EBlJXNJgUFBalBgway2+ljAABAVUOCEAAAAAAAAPBhPP4DAAAAAAAAfBgJQgAAAAAAAMCHkSAEAAAAAAAAfBgJQgAAAAAAAMCH/X/Y8ebG8ytqyAAAAABJRU5ErkJggg==\n",
```
ajlee21, line 124: Neat! So these BRAF scatter plots look as I would expect given the flatness of the curve. However I would've expected the separability to be more muddled in EGFR which actually drops in performance with more data held out (slide 9). Having trouble figuring out how to explain this trend. Maybe this means that the data is nonlinearly separable in some dimension (since the mutated ones are clustering together) but as you hold out more and more you are loosing the signal from the mutated ones. So there seems to be a class imbalance issue here.

03_cross_cancer_classification/plot_holdout_pca.ipynb
```diff
@@ -0,0 +1,385 @@
...
+    "\n",
+    "# calculate the last percent where we have enough data\n",
+    "# (most should be close to 1, but some small cancer types won't be)\n",
+    "last_points = (\n",
+    "    flip_norm_df.groupby(['identifier'])\n",
+    "                .max()['percent_flip']\n",
+    "                .to_dict()\n",
+    ")\n",
+    "\n",
+    "# label cancer types to the right of line, this is much easier to read than colormap\n",
```
ajlee21, line 342: More of a curiosity question: This is also how I have annotated plots in matplotlib. I wonder if there is a cleaner way using pandas (I'm thinking about how the coloring of points works). Looks like maybe ggplot is a little nicer here:) https://ggplot2.tidyverse.org/reference/geom_text.html

ajlee21 (approved):
Really really nice findings! Looks like you're getting closer to solving the mystery!

pancancer_evaluation/utilities/tcga_utilities.py
```diff
@@ -82,7 +82,7 @@ def process_y_matrix(
         suffixes=("_count", "_proportion"),
     ).merge(filter_disease_df, left_index=True, right_index=True)
 
-    if not test:
+    if (not test) and (output_directory is not None):
```
ben-heil, line 85: Technically you don't need parens here because of [logical operator precedence](https://www.mathcs.emory.edu/~valerie/courses/fall10/155/resources/op_precedence.html), but they are useful for grouping so inbd

03_cross_cancer_classification/plot_holdout_pca.ipynb
```diff
@@ -0,0 +1,385 @@
...
+    "    projection_method.upper(), gene, cancer_type))\n",
+    "fig.legend(labels=labels, loc='lower center', ncol=2, fontsize=13)\n",
+    "fig.subplots_adjust(top=0.925, bottom=0.1)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Running the above cell for a few different gene/cancer type combinations shows that in most of the cases where performance is fairly consistent across holdout data percentages (e.g. BRAF_THCA, TP53_LGG) the data is fairly close to being linearly separable. In other words, in these cases data appears to (roughly) cluster by mutation status.\n",
```
ben-heil, line 175: Does separation in UMAP plots actually mean that two classes are linearly separable? Wouldn't it just mean that they're separable in the manifold space? 

I agree that separation = easier classification problem, I'm just asking about whether "linearly separable" is true here since I'm less familiar with the math than you.

ben-heil: Also have you tried running the PCA plots? Do they look similar?

ben-heil (approved):
Very cool! A couple of questions that didn't fit any individual cell:

- How appropriate are linear models in predicting mutations? It like training linear models on multiple cancer types wouldn't necessarily lead to better mutation prediction unless you had really stellar correction across cancer types. Do you expect to use more complex/nonlinear models now that your proof of concept works?
- Is fall off in performance of many of the cancer types' predictors encouraging for the pancancer side of things? Does it indicate that more data is needed in those cancers?
- Is there a way to account for subtypes within the cancer confounding your results? For example in the TP53_LGG plot there are fairly separable groups of TP53 positive and negative cancers, but there is an even bigger divide between what appears to be two types of glioma

jjc2718:
Responses to @ajlee21 comments:

> * Just to clarify, when you say "percent_flipped" = percent held out of training? I assume this is just borrowing old terminology so I just wanted to confirm.

Yeah, sorry, this terminology is left over from when I was actually flipping labels from 1 to 0. I'll update it to `percent_holdout` which I think is clearer.

> * By eye it looks like the trends are similar between the old line plots (proportion of held out) vs new line plots (raw number held out). Can you remind me the logic for this additional plot? I assume that different cancer types will have different numbers of total samples, so I feel like proportion makes more sense to look at in this case. But I'm sure it was mentioned during your BT and I have forgotten. So maybe 95% of the data in cancer A is 400 but cancer B is 100. What does looking at the raw data tell you? I feel like you should just see curves drop off as you move from left to right, but not sure what else...

I think our hypothesis was that the cancer types that don't drop off could be maintaining their performance just because they had more samples. Just to give a slightly more obvious example, if you trained on 5% of a dataset with 5000 samples that's still 250 samples (maybe enough to train a decently generalizable model), but 5% of a dataset with 100 samples is only 5 samples (probably not enough to learn anything generalizable).

Looking at the results, though, your interpretation matches mine - trends look similar to what we had before. It doesn't look like there's much of a difference in the shape of the curves for the cancer types that have lots of samples vs. the ones that have fewer samples, so based on this it seems unlikely that sample size is the main factor behind improved performance in certain cases.

pancancer_evaluation/utilities/tcga_utilities.py
jjc2718: Yeah, I like to add them for logical tests like this because I personally find it a bit clearer (the parens visually set apart the two things I'm trying to compare), but good to know that it isn't strictly necessary.

03_cross_cancer_classification/plot_holdout_pca.ipynb
jjc2718: > Does separation in UMAP plots actually mean that two classes are linearly separable? Wouldn't it just mean that they're separable in the manifold space?

Good question. I had been thinking of it as a supervised vs. unsupervised thing - my gut feeling was that if the labels separate fairly clearly when I apply unsupervised dimension reduction, then a supervised algorithm that's aware of the labels will almost certainly be able to separate them.

To address your point more specifically, I don't think there's any logical/mathematical reason that good UMAP separation in 2 dimensions necessarily implies good logistic regression hyperplane separation in 8000 dimensions - I feel like it would be possible to come up with a counterexample if you thought hard about it. But my suspicion is that this sort of thing would be uncommon in practical situations (although I haven't tested this).

I think if I was going to use this approach more generally to identify easy/hard classification problems I'd probably simulate some data to understand the tradeoffs a bit better, but for purposes of what I'm doing here I think it's probably fine to assume UMAP separation => decent classification in most cases (although I don't know if the reverse is true, even in practice).

> Also have you tried running the PCA plots? Do they look similar?

I have - for the most part they do look pretty similar. The "clusters" are a bit less obvious (higher variance), but the trends seem to be the same.

jjc2718:
Responses to @ben-heil comments:

> * How appropriate are linear models in predicting mutations? It like training linear models on multiple cancer types wouldn't necessarily lead to better mutation prediction unless you had really stellar correction across cancer types. Do you expect to use more complex/nonlinear models now that your proof of concept works?

Yeah, it's possible that using more complex models would change things. [This paper](https://doi.org/10.1093/bib/bbaa258) was published recently showing that non-linear models can work well for predicting Ras pathway activation - although I'm not entirely convinced that their particular method will generalize well outside of Ras (a relatively easy example with lots of training data), I think there's enough evidence there that the general idea might be worth trying with a more established non-linear model.

> * Is fall off in performance of many of the cancer types' predictors encouraging for the pancancer side of things? Does it indicate that more data is needed in those cancers?

I definitely think these results do indicate that more data would be beneficial in some cases, which is encouraging. I just think it's unclear whether pan-cancer data necessarily solves the problem in all cases, or whether the differences between cancer types overwhelm the benefits of having more data (especially in light of our results from the pan-cancer vs. single-cancer experiments from before). Interested in continuing to think about this going forward!

> * Is there a way to account for subtypes within the cancer confounding your results? For example in the TP53_LGG plot there are fairly separable groups of TP53 positive and negative cancers, but there is an even bigger divide between what appears to be two types of glioma

Good point. The two types of glioma we're seeing are almost certainly IDH1 mutants vs. non-mutants (like we saw in greenelab/mpmp#2). TCGA does have some subtype annotations for certain cancers (IDH1 status in glioma is among them), so using these as a covariate may be something worth trying in the future. But some of these subtypes are poorly defined or hotly debated, like Ariel's work has shown us in the case of ovarian cancer, so I don't have a good sense of how helpful they'll actually be or which cases they might be helpful in.

03_cross_cancer_classification/plot_holdout_pca.ipynb
jjc2718: Yeah, I think this is kinda the accepted way to do this in matplotlib (at least from what I've read on Stack Overflow) but it would be nice if there was a cleaner way.

I also recently came across `plt.annotate` [(docs here)](https://matplotlib.org/3.3.3/api/_as_gen/matplotlib.pyplot.annotate.html), which seems like it could be a bit cleaner than writing a function/`for` loop. There's also [this Python library](https://github.com/Phlya/adjustText) to do a similar non-overlapping text adjustment as ggplot/ggrepel.

I should really set aside some time to try these things out, but for purposes of this plot I just wanted to get something down that was somewhat legible, so I just copied some of my old code and modified it a bit. Will definitely think about neater approaches in the future!

03_cross_cancer_classification/plot_holdout_pca.ipynb
jjc2718: Yeah, I think there are a few things to note (but I'm guessing a bit at this point):

* Maybe the relationship between features and labels is complex/multi-dimensional, and not really clearly visible in the 2D UMAP plots (but the model can find it). EGFR in lung cancer is a particularly interesting case since it's a pretty well-known biomarker that correlates with response to certain targeted therapies (see, e.g. [this paper](https://doi.org/10.1126/science.1099314)), so I expected there to be some relationship between EGFR mutation status and the gene expression data, but it's not really as strong as I expected across holdout sets.
* Maybe the label imbalance is driving the decrease in performance, like you're suggesting (even though there's still a decent amount of training data overall in the 95% case, there are very few positive labels so it may be hard to train a good model)
* Maybe we're just sampling bad training sets (but I doubt it, results across 3 random seeds were pretty similar)

If I had to guess I'd say what we're seeing represents a combination of all of the above, but I'm not sure!

03_cross_cancer_classification/plot_holdout_pca.ipynb
ajlee21: This is a great point. Its been a while since looking at the umap paper, but my intuition is that if the labels were linearly separable, I'd expect to see a pretty clear separation in UMAP. If the labels were nonlinearly separable, I'd expect that you might see the separation in UMAP but not necessarily...I'd have to check how the math works though...just my intuition

PR #42, opened 2021-01-14
Add cancer types to single-cancer mutation prediction models

jjc2718:
The code changes for this PR aren't too extensive, but the results might be a bit dense/hard to parse. Let me know if you have questions on them, happy to clarify things.

**Overview of PR goals:**

As a reminder, in past experiments (#21 and others) I showed that training on a pan-cancer dataset doesn't improve mutation prediction in most cases compared to training only on the cancer type in the test set.

In the experiments in this PR, we hypothesized that we could do better than a pan-cancer model by adding only a subset of cancer types that are relevant to the target cancer. Our guess/hope was that in the pan-cancer data, some of the cancer types would be unrelated or have different mutational consequences than the target cancer, which could muddle the signal gained by adding more data.

Our general idea was to take the infrastructure from previous experiments (predicting mutations in a given gene for a given cancer type), and instead of trying only single-cancer or pan-cancer models, we iteratively add other cancer types to the training set to see if performance improves or gets worse.

To choose cancer types that are "similar" to the target cancer types, we used our confusion matrices from the experiments here: https://github.com/greenelab/mpmp/pull/13. We just used the gene expression matrix (subsampled version) for now.

**Results:**

You can see results for a variety of genes and cancer types here: https://docs.google.com/presentation/d/1Zd5bnUrTEdZDoPVFlCH5Yt1hcguUefv05_pWCNRh_IY/edit?usp=sharing . Unfortunately, overall adding cancer types by confusion matrix similarity doesn't seem to make a drastic difference in performance in most cases. There are a few interesting cases where it does change performance, but not as many as we were expecting/hoping for.

Example here for TP53:

image

Most cancer types are linear as more data is added (little difference between single-cancer and pancancer models). Some have a slight upward slope (pancancer model helps), but few are "peaky" in the middle (adding a few cancer types helps but pancancer model is worse), which is what we were hoping to see.


pancancer_evaluation/exceptions.py
```diff
@@ -10,8 +10,7 @@ class NoTrainSamplesError(Exception):
     This allows calling scripts to choose how to handle this case (e.g. to
     print an error message and continue, or to abort execution).
     """
-    def __init__(self, *args):
-        super().__init__(*args)
+    pass
```
ben-heil, line 13: Just curious, why did you convert the init functions for your exceptions? Is the superclass constructor called on init by default? Or does it not matter because you aren't passing messages or something?

ben-heil (approved):
I forget, have you tested whether the AUPR decreases when you only use a subset of your given cancer type? If the cancer types were similar enough I'd expect the AUPR to increase just by virtue of having more data. 

My hypotheses for the flat prediction curves:
1. You're running into irreducible error. Mutations only have so much impact on gene expression so the inherent noise makes it impossible to improve past the point you get to with your starting number of samples
2. Linear models are insufficiently complex. Basically the same as 1, but potentially you could shift the curve up if you were able to introduce more complicated decision boundaries
3. Maybe some flavor of [gradient starvation](https://arxiv.org/abs/2011.09468)? I wouldn't expect that to be the case on a linear model, but maybe your model has an easy shortcut to learning the training data so it misses some of the nuance that would be helpful?
4. Predictions from gene expression data aren't actually possible. In that case maybe your model learns some technical artifact, but the amount of biological signal you bring to the table doesn't really matter. I don't think that's true, but it's worth thinking about
5. Applying Occam's Razor to 3 rather than invoking trendy DL papers, maybe the models are just learning detecting technical artifacts before biology

pancancer_evaluation/data_models/tcga_data_model.py
```diff
@@ -509,3 +637,77 @@ def holdout_percent_labels(y,
...
+                                 if ct != test_cancer_type]
+
+        if num_cancer_types == -1:
+            # pan-cancer model, train on all valid cancer types
+            train_cancer_types += valid_cancer_types
+        elif num_cancer_types >= 1:
+            # add desired number of cancer types to train set
+            if how_to_add == 'random':
+                import contextlib
+                # We want this random addition of cancer types to be the same
```
ajlee21, line 661: I'm not sure if i understand the logic for why you're re-seeding here. So each time you're adding a random cancer type so that you have target + A or target + A,B or target + A, B, C. And each time you want to preserve the previously added cancers + adding a new random one. Why doesn't using a global see and say passing in a seed to select the additional one cancer work?

ajlee21 (approved):
I think Ben has made some really good points. To add onto that:

1. To clarify, you selected additional cancer types that your confusion matrix suggested have similar gene expression signatures as your training/target cancer types. And so you're hoping that the added data would boost your signal? I guess if you're just adding more of the same signal I'm not sure how much you would improve in theory (I think this is similar to Ben's (1)). I'm curious if you would gain additional rare/complex signals using a more complex model as Ben already mentioned too (2)
2. Interesting behavior by some gene/cancer types in your slides. I'm wondering about the peak in BRAF_COAD (slide 3). And the case where you're seeing a steady increase for NF1 (slide  11). What do you think is contributing to the increase in performance in these cases? So is your next step to determine if these cherry-picked example trends are robust?

jjc2718:
Thanks for the feedback @ben-heil! These are good ideas.

> I forget, have you tested whether the AUPR decreases when you only use a subset of your given cancer type? If the cancer types were similar enough I'd expect the AUPR to increase just by virtue of having more data.

This is sort of what we were doing in #37. Just as a reminder, here are the results we got for progressively larger holdout sets (i.e. progressively smaller training sets) within each cancer type:

image

You can read these a bit like "reverse learning curves" (i.e. larger x-axis => less training data), but the size of the holdout set is changing as more data is held out, so it might be good to hold that constant and see what happens with smaller training sets.

> My hypotheses for the flat prediction curves:
> 
> 1. You're running into irreducible error. Mutations only have so much impact on gene expression so the inherent noise makes it impossible to improve past the point you get to with your starting number of samples

Yeah, I think this is definitely plausible. Definitely for some of these examples, the mutation probably doesn't affect gene expression at all (i.e. it's not a driver in that cancer type), so that would explain a flat curve. We were just surprised to see so many examples where this is the case.

> 2. Linear models are insufficiently complex. Basically the same as 1, but potentially you could shift the curve up if you were able to introduce more complicated decision boundaries

By "shift the curve up", you mean better performance for all training set sizes? Or more improvement as training data increases? Or both?

I think it's possible that this would help...will have to think about it.

> 3. Maybe some flavor of [gradient starvation](https://arxiv.org/abs/2011.09468)? I wouldn't expect that to be the case on a linear model, but maybe your model has an easy shortcut to learning the training data so it misses some of the nuance that would be helpful?

This is interesting - I hadn't seen the paper, but I do think based on some experiments that I ran in the past, our models' feature selection is a bit unstable (e.g. between bootstrapped subsets of the training data, nonzero coefficients vary a lot). So it's possible that better choices of features would result in models that generalize better.

I'm not sure exactly what I would try first as far as addressing this (there are tons of feature selection methods we haven't looked at), but definitely an idea for the future.

> 4. Predictions from gene expression data aren't actually possible. In that case maybe your model learns some technical artifact, but the amount of biological signal you bring to the table doesn't really matter. I don't think that's true, but it's worth thinking about
> 5. Applying Occam's Razor to 3 rather than invoking trendy DL papers, maybe the models are just learning detecting technical artifacts before biology

Yeah, these are good points. I don't have any great ideas for how to separate technical artifacts from actual biological signal, but maybe I'll have to do some reading. Interested to hear if you have any ideas!

pancancer_evaluation/exceptions.py
jjc2718: Based on [this stackoverflow thread](https://stackoverflow.com/a/1319675), it seems like if all you're doing is passing messages you don't need to call the constructor manually or override anything. I'm not exactly sure how that works internally, but I definitely see the messages when I call my custom exceptions (i.e. the init call didn't seem to be adding anything).

jjc2718:
Responses to @ajlee21 comments:

> 1. To clarify, you selected additional cancer types that your confusion matrix suggested have similar gene expression signatures as your training/target cancer types. And so you're hoping that the added data would boost your signal? I guess if you're just adding more of the same signal I'm not sure how much you would improve in theory (I think this is similar to Ben's (1)). I'm curious if you would gain additional rare/complex signals using a more complex model as Ben already mentioned too (2)

Yeah, what you said sounds exactly right. Particularly for small sample size cancers, we were hoping that adding more data (where the data is from a cancer type similar to the target cancer type) would improve the predictions. I do think it would be good to try to subset some of these cancers and see if the model seems to be saturated (kind of like some of the experiments Ben has done in his work), which may give us a better idea of how fruitful adding more data is likely to be in the best case.

> 2. Interesting behavior by some gene/cancer types in your slides. I'm wondering about the peak in BRAF_COAD (slide 3). And the case where you're seeing a steady increase for NF1 (slide  11). What do you think is contributing to the increase in performance in these cases? So is your next step to determine if these cherry-picked example trends are robust?

Yeah, I think these are interesting cases. BRAF mutations in colon cancer seem to be a bit of an odd case (they're different in some way than in other cancers), and uterine cancer (added to form the peak) may just be more similar than other cancers (in other words, this could be a real interesting signal). Not so sure about NF1.

I think we may eventually look into whether these trends are robust, but we were hoping to see a stronger signal across the board (i.e. this was a bit of a pilot study before trying more genes and more elaborate methods), so I'm not sure how much more time I'll spend pursuing this given the lukewarm outcome. Even if this did work well, it's unclear how useful it would be, as compared to (for example) the methylation project where the signal/takeaway seems to be a bit clearer.

pancancer_evaluation/data_models/tcga_data_model.py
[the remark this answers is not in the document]
jjc2718: This function can be run multiple times by the script that calls it (`run_add_cancer_classification.py`). Each time I call this function, I want to choose cancers randomly in the same order (like you wrote above), so I want to reset the seed locally. But I don't want to reset the global numpy seed each time, since that might affect other things going on in the script that calls this code (e.g. cross-validation splits, model optimization, etc). Does that answer your question?

jjc2718: I think this is a duplicate of your first comment (let me know if I'm wrong).

ajlee21: yup..i was moving around tabs and refreshing things so you can just ignore this. I thought the comment was lost when i closed the tab

04_add_cancer_types/run_add_cancer_classification.py
```diff
@@ -0,0 +1,257 @@
...
+    NoTrainSamplesError,
+    NoTestSamplesError,
+    OneClassError,
+    ResultsFileExistsError
+)
+from pancancer_evaluation.utilities.classify_utilities import run_cv_cancer_type
+import pancancer_evaluation.utilities.data_utilities as du
+import pancancer_evaluation.utilities.file_utilities as fu
+
+def process_args():
```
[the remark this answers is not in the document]
jjc2718: > Just curious to learn more here -- so this allows you to run this script from the terminal? Is there a reason you don't call this from a notebook? 
 
My main reason is just personal preference. For this script, I want to rerun it multiple times with a few different parameters (i.e. different gene sets, different cancer type addition methods), and it takes a while to run. I think both of these things are doable in notebooks: I know Milton uses [Jupytext](https://jupytext.readthedocs.io/en/latest/index.html) to run his notebooks as scripts with different parameters, and I think in theory long running jobs should work fine in notebooks. But I have more experience using Python scripts for these types of tasks so I stick with that for now.

I like notebooks a lot for doing quick, iterative analysis and visualization of results (I often end up making lots of small changes to plots, and it's really tedious to have to rerun a script and open an image file each time I make a change, as opposed to seeing the changes instantly inline in a notebook).

PR #45, opened 2022-06-09
Feature stability experiments

jjc2718:
Reviving this repo a bit - a while ago, we observed that mutation prediction models trained on the same dataset had very different coefficients. We wanted to look at how this affected performance, both within the training set (cross-validation folds) and across cancer types.

For now, we're just looking at TP53, across the cancer types with the most mutated samples. In general, we see a decrease in performance across cancer types (as expected), and also an increase in variance.

The latter observation means that even for models that perform similarly on the training set, we can sometimes see very different capacity to generalize to different datasets. We'll probably explore some options/hypotheses for why that could be the case next.

image


03_cross_cancer_classification/nbconverted/p53_stability_pilot.py
```diff
@@ -0,0 +1,186 @@
...
+heatmap_df.iloc[:5, :5]
+
+
+# In[5]:
+
+
+sns.set({'figure.figsize': (15, 8)})
+sns.heatmap(heatmap_df, cbar_kws={'label': 'AUPR difference from baseline'}, center=0)
+plt.title('Cross-cancer mutation detection, AUPR heatmap')
+plt.xlabel('Test identifier')
```
arielah, line 60: Is there something about UCEC that makes it easier to predict? It seems to have overall better performance. Maybe I'm reading too much into the heatmap.

03_cross_cancer_classification/nbconverted/p53_stability_pilot.py
```diff
@@ -0,0 +1,186 @@
...
+
+num_ids = len(plot_big_df.train_identifier.unique())
+for row in range(num_ids):
+    diag_ix = (num_ids * row) + row
+    g.axes[diag_ix].set_facecolor('xkcd:light blue grey')
+
+
+# In[10]:
+
+
```
arielah, line 153: The axis labels on the grid aren't really visible (at least not on the github website), don't know if that's a dealbreaker.

03_cross_cancer_classification/nbconverted/p53_stability_pilot.py
```diff
@@ -0,0 +1,186 @@
...
+g.set_titles(col_template='{col_name}')
+num_ids = len(plot_big_df.train_identifier.unique())
+for row in range(num_ids):
+    diag_ix = (num_ids * row) + row
+    g.axes[diag_ix].set_facecolor('xkcd:light blue grey')
+
+
+# In[11]:
+
+
```
arielah, line 168: Could benefit from a header or a comment explaining what's different about this plot and the one in [10]

arielah (approved):
Looks good! I'm interested to see what else comes of this part of the project.

.gitignore
```diff
@@ -8,4 +8,4 @@ data/EBPlusPlusAdjustPANCAN_IlluminaHiSeq_RNASeqV2-v2.geneExp.tsv
 data/*.gz
 data/*.pkl
 
-results/*
+**/results/*
```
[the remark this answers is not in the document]
jjc2718: I don't feel too strongly either way. I feel like it doesn't hurt, which is why I do it, but I don't think you need to.

03_cross_cancer_classification/nbconverted/p53_stability_pilot.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# ## p53 feature stability analysis
```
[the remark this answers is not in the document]
jjc2718: Yeah, basically as a pilot before eventually generalizing to other genes.

p53 is a bit unique in that it's the one gene where we saw the models transfer across domains (tissues) pretty well in general - other genes we tried before (KRAS, BRAF, etc) didn't work quite as well. That said, we're mostly interested in the variance across domains here (not the overall performance) so we'll probably try some others to see if that result is the same.

03_cross_cancer_classification/nbconverted/p53_stability_pilot.py
jjc2718: I'm not too worried about it at this point. If we end up using these plots for anything else in the future I'll clean up the formatting.

The x-axis ticks are train/test and the y-axis is performance (AUPR), though, in case you're wondering.

03_cross_cancer_classification/nbconverted/p53_stability_pilot.py
jjc2718: I'm not sure! I'm not surprised that SARC and LGG (the two non-carcinoma cancer types) are generally harder to predict than the others, but I'm not sure why UCEC would be easier to predict than the rest of the carcinomas.

Maybe it's that the labels are a bit more imbalanced (around 35% mutated as opposed to closer to 50% for most of the other cancer types)? BRCA has around the same % mutated as UCEC, though, and performance isn't as good.

PR #47, opened 2022-07-21
Univariate feature selection for stratified CV

jjc2718:
We're testing an idea that features that "generalize better" across cancer types on the training set, for some definition of "generalize better", will allow us to develop models that are more robust to shifts in cancer type/tissue etc.

This experiment is sort-of a sanity check, since we're just doing stratified cross-validation (i.e. every cancer type in the test set is also in the training set). We mostly want to see that the new feature selection methods don't completely tank performance in this case, before moving on to the case where we're looking at generalization to specific cancer types, that may or may not be present in the training data.

We do seem to see that performance is fine for the new methods ("pancan_f_test" and "median_f_test" are the two that should be roughly close to the green box, which is no feature selection here):

image


01_stratified_classification/nbconverted/plot_univariate_fs_results.py
```diff
@@ -0,0 +1,335 @@
...
+# ## Analysis of feature selection results
+# 
+# We tried a few different feature selection methods for mutation prediction:
+# 
+# * mean absolute deviation
+# * pan-cancer univariate f-statistic (on training data, separate samples into mutated/not mutated and do an f-test)
+# * median univariate f-statistic across cancer types (same as above for each individual cancer type, then take median)
+# * MAD of univariate f-statistic across cancer types (same as above, but look for least variable genes)
+# 
+# Here, we're just looking at stratified CV across all cancer types, as a sanity check - the idea is to make sure the univariate correlation-based feature selection methods perform reasonably well. This isn't necessarily where we'd expect them to help performance, but we still don't want them to _hurt_ it.
```
arielah, line 13: Very good intro, helps me (and probably future you) know exactly what the motivation is.

01_stratified_classification/nbconverted/plot_univariate_fs_results.py
```diff
@@ -0,0 +1,335 @@
...
+get_ipython().run_line_magic('load_ext', 'autoreload')
+get_ipython().run_line_magic('autoreload', '2')
+
+
+# ### Load results and look at overall performance
+
+# In[2]:
+
+
+results_dir = os.path.join('results', 'univariate_fs', 'pancancer')
```
arielah, line 40: In this section maybe a quick "looking at results that got generated by this script" pointer

01_stratified_classification/nbconverted/plot_univariate_fs_results.py
```diff
@@ -0,0 +1,335 @@
...
+    plot_df = compare_df[compare_df.identifier == gene]
+    sns.boxplot(data=plot_df, x='fs_method', y='delta_aupr', ax=ax)
+    ax.set_title(gene)
+    ax.set_xlabel('Feature selection method')
+    ax.set_ylim(0, 1)
+
+plt.tight_layout()
+
+
+# We can see that the blue box (median f-test) and the purple box (pan-cancer f-test) are reasonably similar in most cases to the green box (no feature selection from 1000 features), suggesting that selecting features by f-statistic doesn't seem to harm performance, at the very least. This is good.
```
arielah, line 99: Did you pick these six genes for a reason?

01_stratified_classification/nbconverted/plot_univariate_fs_results.py
```diff
@@ -0,0 +1,335 @@
...
+        id_coefs_info.append([identifier, fold_no, coefs])
+        
+print(len(id_coefs_info))
+
+
+# In[7]:
+
+
+print(len(id_coefs_info[0]))
+print(id_coefs_info[0])
```
arielah, line 127: This is not interpretable by eye. If that's okay, then no worries, but if it is important, maybe a couple lines of code could format it into a table?

01_stratified_classification/nbconverted/plot_univariate_fs_results.py
```diff
@@ -0,0 +1,335 @@
...
+upset_series = series_from_samples(
+    list(fs_method_coefs.values()), list(fs_method_coefs.keys())
+)
+upset_series[upset_series != 0].sort_values(ascending=False).head(5)
+
+
+# In[12]:
+
+
+up.plot(upset_series[upset_series != 0])
```
arielah, line 191: Making sure I'm keeping track of this. This is showing the number of cancer genes where TP53 is selected as a useful feature for prediction by all these various feature selection methods?

01_stratified_classification/nbconverted/plot_univariate_fs_results.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+y_df = (mutation_df
+    .loc[:, [gene]]
+    .merge(sample_freeze_df, left_index=True, right_on='SAMPLE_BARCODE')
+    .drop(columns='PATIENT_BARCODE')
+    .set_index('SAMPLE_BARCODE')
+    .rename(columns={gene: 'status',
+                     'DISEASE': 'cancer_type',
+                     'SUBTYPE': 'subtype'})
```
arielah, line 264: Is the subtype being used in the model? Because I imagine you don't have that info for all cancer types and it seems like that could really affect the model...

pancancer_evaluation/utilities/feature_utilities.py
```diff
@@ -0,0 +1,91 @@
...
+    X_test_df = X_test_df.reindex(valid_features, axis='columns')
+    gene_features = np.concatenate((
+        np.ones(top_features.shape[0]).astype('bool'),
+        np.zeros(non_gene_features.shape[0]).astype('bool')
+    ))
+
+    return X_train_df, X_test_df, gene_features
+
+
+def _get_cancer_type_f_statistics(X_train_df, y_df):
```
arielah, line 47: This may be blatantly obvious to anyone with more Python training than me, but what's the significance of putting the underscore in front of some functions but not others? You're not planning to call those directly in your main scripts?

tests/test_feature_weights.py
```diff
@@ -0,0 +1,46 @@
+"""
+Test cases for feature_utilities.py
+"""
+import pytest
+import numpy as np
+import pandas as pd
+
+import pancancer_evaluation.utilities.feature_utilities as feats
+
```
arielah, line 9: Why does this file look nothing like test_feature_weight.ipynb? Am I missing something?

arielah (approved):
Looks good! It's a lot tidier than a lot of my exploratory analyses but I suppose that's to be expected from PR#47. :)

tests/test_feature_weights.py
jjc2718: Good catch! `test_feature_weight.ipynb` was a notebook I wrote to try out some testing methods, which I eventually made into `pytest` unit tests in the `tests` directory. So they're really pretty different files, not intended to just be a converted notebook like the other ones.

I actually didn't mean to commit `test_feature_weight.ipynb` - I'll delete it and leave the unit tests.

pancancer_evaluation/utilities/feature_utilities.py
jjc2718: Yeah, it's the Python "internal method" convention (Python doesn't really support private/internal methods, so it's more like "please don't use this function outside of this file, even though you technically can"). I mostly use it when I want to write a complicated helper function but not clean it up or make it general enough to be used elsewhere, haha.

More detail here, if you're interested: https://stackoverflow.com/a/6930223

01_stratified_classification/nbconverted/plot_univariate_fs_results.py
jjc2718: Currently we're using it in the train/test set stratification - in other words, if a cancer type has subtypes, we try to balance the subtypes between train/test sets so that proportions are relatively equal. For cancer types that don't have well-defined subtypes we just ignore it and stratify by the cancer type as a whole.

We're not actually using it for prediction (i.e. as a feature in the model), though, for the reason you suggested.

01_stratified_classification/run_univariate_fs.sh
```diff
@@ -0,0 +1,80 @@
...
+genes=(
+  # "TP53"
+  "APC"
+  "EGFR"
+  "SETD2"
+  "ARID1A"
+  "PIK3CA"
+)
+
+for seed in 42 1; do
```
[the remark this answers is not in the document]
jjc2718: Yeah, I should probably just give it a try at this point. I will consider it for future PRs!

01_stratified_classification/nbconverted/plot_univariate_fs_results.py
jjc2718: (I'll put this in the script as a comment also)

01_stratified_classification/nbconverted/plot_univariate_fs_results.py
jjc2718: I'll try to make this more interpretable - the idea of this cell is just to show what the format of the list is, but I agree it's a bit of a wall of text and colud be clearer.

01_stratified_classification/nbconverted/plot_univariate_fs_results.py
jjc2718: Not quite, this is a bit confusing for sure - for models where we're predicting TP53 mutation status with various feature selection methods, this is showing the overlap in the gene expression features they selected (either the unique features for the single dots, or shared features for the connected dots in the upset plot). I'll add this to the Markdown cell too.

PR #48, opened 2022-08-01
Univariate feature selection, held-out cancer types

jjc2718:
Really this should probably be two separate PRs -- the added analysis in `01_stratified_classification/nbconverted/plot_univariate_f_dists.py` and the changes in `02_cancer_type_classification` -- sorry in advance.

The main changes in this PR are the modifications to `02_cancer_type_classification/run_cancer_type_classification.py`, allowing us to test the same feature selection methods described in #47 when entire cancer types are held out of the training set, and the analysis of those results is in `02_cancer_type_classification/plot_univariate_fs_results.ipynb`.

We're planning to look across more genes in the future, but for TP53 and EGFR, the results are pretty close to what we expected: for cancer types that are not very related to the training set, selecting by median correlation across cancer types (i.e. selecting features that are "generally" predictive) improves generalization performance vs. selecting by aggregate pan-cancer correlation (which can choose features that are "specifically" predictive or driven strongly by one/a few cancer types).

Here's an example: compare the "train pan-cancer" (test cancer present in training set) and "train all other cancers" (test cancer left out of training set) for EGFR. We can see that `pancan_f_test` (green box) performance decreases for the latter, and `median_f_test` (red box) is more resilient to the test cancer type being dropped out.

image

PIK3CA, however, doesn't really follow this pattern:

image


01_stratified_classification/nbconverted/plot_univariate_fs_results.py
```diff
@@ -40,7 +40,14 @@
 # analysis of results generated by script:
 # 01_stratified_classification/run_stratified_classification.py
 # (with varying feature_selection parameters)
-results_dir = os.path.join('results', 'univariate_fs', 'pancancer')
+
+# results_dir = os.path.join('results', 'univariate_fs', 'pancancer')
```
arielah, line 44: Do you still need this code?

pancancer_evaluation/utilities/data_utilities.py
```diff
@@ -258,9 +258,17 @@ def split_by_cancer_type(rnaseq_df,
     """Split expression data into train and test sets.
 
     The test set will contain data from a single cancer type. The train set
-    will contain either the remaining data from that cancer type, or the
-    remaining data from that cancer type and data from all other cancer types
-    in the dataset.
+    will contain one of the following:
+        - only the remaining data from the same cancer type
```
arielah, line 262: This is a good explanation of what's going on. Is there an argument for making it one variable with three levels? ie just_cancer_type, just_other_type, all?

01_stratified_classification/nbconverted/plot_univariate_f_dists.py
```diff
@@ -0,0 +1,337 @@
...
+        id_coefs_info.append([identifier, fold_no, coefs])
+        
+print(len(id_coefs_info))
+
+
+# In[4]:
+
+
+print(len(id_coefs_info[0]))
+print(id_coefs_info[0])
```
arielah, line 52: Maybe a quick comment about what someone should take from this? Just that it's properly formatted?

01_stratified_classification/nbconverted/plot_univariate_f_dists.py
```diff
@@ -0,0 +1,337 @@
...
+fig, axarr = plt.subplots(2, 3)
+
+genes_to_plot = mad_df.index[:6]
+for ix, gene_to_plot in enumerate(genes_to_plot):
+    ax = axarr[ix // 3, ix % 3]
+    dist_df = f_df.loc[gene_to_plot, :]
+    sns.histplot(dist_df.values, ax=ax, bins=20)
+    ax.set_title(gene_to_plot)
+    
+plt.suptitle('Top MAD genes')
```
arielah, line 335: What's the y axis here?

02_cancer_type_classification/run_cancer_type_classification.py
```diff
@@ -49,6 +54,12 @@ def process_args():
                         'cancer types in TCGA if none are provided')
     p.add_argument('--log_file', default=None,
                    help='name of file to log skipped cancer types to')
+    p.add_argument('--mad_preselect', type=int, default=None,
+                   help='if included, pre-select this many features by MAD, '
+                        'before applying primary feature selection method')
+    p.add_argument('--num_features', type=int, default=cfg.num_features_raw,
```
arielah, line 60: What's the difference between mad_preselect and num_features? Their description reads the same to me.

02_cancer_type_classification/nbconverted/plot_univariate_fs_results.py
```diff
@@ -0,0 +1,377 @@
...
+                order=fs_method_order, ax=ax)
+    ax.set_title(names_to_plot[ix])
+    ax.set_xlabel('Feature selection method')
+    ax.set_ylim(-0.2, 1)
+    
+plt.suptitle('{}, averaged over non-carcinoma cancer types'.format(gene))
+plt.tight_layout()
+
+
+# ### Plot performance broken down by cancer type
```
arielah, line 283: I think I got confused here. Are you doing a pairwise experiment? I assumed "all other cancer types" was all of them concatenated, not one at a time, so I don't know how the plots in chunk 12 work.

arielah (approved):
This looks good! I got a little turned around in the middle of `plot_univariate_fs_results.py` but by the end I think I understood what you were trying to do. Interesting analysis! This seems like a useful result for the field.

01_stratified_classification/run_univariate_fs.sh
```diff
@@ -2,13 +2,20 @@
 
 # Run feature selection experiments across a few genes
 
-RESULTS_DIR=./01_stratified_classification/results/univariate_fs
+RESULTS_DIR=./01_stratified_classification/results/univariate_fs_big
 ERRORS_DIR=./stratified_fs_errors
 
+# number of features to use as 'upper bound' model by variance selection
+# this should be larger than the actual experimental methods
+NUM_FEATURES_LARGE=-1
```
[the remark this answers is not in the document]
jjc2718: Right, I'll add a comment here: -1 means no feature selection (so `NUM_FEATURES_LARGE` would actually be 16,000 or however many gene expression features there are).

In retrospect probably `None` or something like that would have been clearer, but for now I'll just add a note.

01_stratified_classification/nbconverted/plot_univariate_fs_results.py
jjc2718: Nope, will remove.

02_cancer_type_classification/run_cancer_type_classification.py
jjc2718: `mad_preselect` happens _before_ `feature_selection`/`num_features`. It's mainly a way to speed things up a bit for our early experiments here - since calculating the f-statistic for each of 16,000 features in each cancer type can be slow, sometimes it makes sense to start with (for example) the top 5,000 most variable features, then do the univariate correlation calculations only on those features. I'll try to make this clearer in the help strings.

pancancer_evaluation/utilities/data_utilities.py
jjc2718: Yeah, that would make more sense! In the scripts I'm running directly (e.g. `run_cancer_type_classification.py`) it makes sense to have the `pancancer_only` argument (although that's not a great name), but for this function I think you're right, it would be clearer to have one variable with three levels.

02_cancer_type_classification/nbconverted/plot_univariate_fs_results.py
jjc2718: The training set is all of them concatenated, but the test set for the "cancer type" experiments in this script is always only a single cancer type, as opposed to the "stratified" experiments from before where the test set is also all cancer types concatenated.

This plot is showing performance on the test set, so we can split that out by cancer type even though the training set is (largely) the same data.

01_stratified_classification/nbconverted/plot_univariate_f_dists.py
jjc2718: Will clarify!

PR #49, opened 2022-08-04
Add more genes and random features baseline to cancer type experiments

jjc2718:
As a continuation of #48, we ran the same experiments for more genes, and added random feature selection (of the same number of features) as a baseline.

General results are similar to what we saw before - PTEN is an example of a gene where our feature selection method (`median_f_test`) works well for non-carcinoma cancer types:

image

Some other genes we tried (RB1, BRAF) don't seem to work as well, though.

I also added a script at `02_cancer_type_classification/scripts/run_fs_for_genes.sh` to run `02_cancer_type_classification/plot_univariate_fs_results.ipynb` for all the genes we looked at and save the resulting plots (I didn't track the plots here, just locally).


02_cancer_type_classification/run_cancer_type_classification.py
```diff
@@ -42,7 +42,7 @@ def process_args():
     p.add_argument('--debug', action='store_true',
                    help='use subset of data for fast debugging')
     p.add_argument('--feature_selection',
-                   choices=['mad', 'pancan_f_test', 'median_f_test', 'mad_f_test'],
+                   choices=['mad', 'pancan_f_test', 'median_f_test', 'random'],
```
arielah, line 45: I get that you're adding random feature selection, but can you explain why you're getting rid of the mad_f_test?

arielah (approved):
This all looks fine to me!

02_cancer_type_classification/run_cancer_type_classification.py
jjc2718: `mad_f_test` didn't really make any sense to begin with - the idea was to select features that have low variability across cancer types, but what that ends up doing is just picking features that aren't correlated with the label in any cancer type, since if a feature has ~0 correlation in all cancer types it will have low variance.

I had been including `mad_f_test` as a baseline/lower bound, but just selecting random features pretty much always outperforms it, and that seems like a more sensible baseline to use going forward.

PR #50, opened 2022-08-26
Purity classification (above/below median)

jjc2718:
Adding classification of tumor purity above/below the median, for each cancer type, as another problem to test our "median across cancer types" feature selection approach on.

Results are somewhat similar, but less drastic than, some of the mutation prediction examples we looked at in #49. Here are the results for non-carcinoma cancer types (where we expect `median_f_test` to work well):

image

So we can see that `median_f_test` does perform the best in general, but it's not a huge effect, and the variance across cancer types is pretty large.


07_purity_prediction/download_purity.ipynb
```diff
@@ -0,0 +1,658 @@
+{
```
ben-heil, line 1: <p>Is there a reason you chose a stripplot here over a violin plot? Is the purpose to show the differences in ranges of the different cancer types?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/50/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

07_purity_prediction/feature_robustness.ipynb
```diff
@@ -0,0 +1,516 @@
+{
```
ben-heil, line 1: <p>Are these results what you expected? I guess I'm not sure what I'd expect the distribution to look like ahead of time, but isn't it interesting that genes are either always useful or never useful regardless of cancer type? </p><p> </p><p>Doesn't this imply that in the pancancer f-test feature selection ~40% of your features could be thrown out and still see the same performance?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/50/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

07_purity_prediction/download_purity.ipynb
jjc2718: <p>I think either would be fine. I just generally wanted to see if any of the purity distributions are particularly skewed or if any cancer types are drastically under-represented compared to the others (the latter seems a bit easier to see with a strip plot, although I guess I could make a violin plot with the sample sizes for each cancer type as labels). This was just an exploratory visualization, though, so I don't want to spend too much time tweaking it right now.</p>

07_purity_prediction/feature_robustness.ipynb
jjc2718: <blockquote>Are these results what you expected?</blockquote><p> </p><p>Yeah, these are more or less what I expected. These results are for the "all other cancers" cross-validation splits, which means that if we're testing on cancer type X, we're training on all other cancer types in TCGA minus X. Since there are 33 cancer types in TCGA, this means the training set isn't changing too much between cancer types and cross-validation folds, so I'd expect the features selected to be pretty consistent for all the non-random selection methods.</p><p> </p><p>At some point I should probably do this for the single cancer cross-validation splits, where I would expect the features selected to vary quite a bit more because the training sets are completely different for each cancer type.</p><p> </p><blockquote>Doesn't this imply that in the pancancer f-test feature selection ~40% of your features could be thrown out and still see the same performance?</blockquote><p> </p><p>I guess so, but it's hard to know a priori which 40%. We already know from past experiments (and other studies have shown the same thing) that many small subsets of features can achieve similar performance on the same dataset for gene expression, since there's so much redundancy/correlation in the data.</p><p> </p><p>Our goal with this study is to determine if there's a simple way to select "general" features -- i.e. from the redundant ones, select ones that are more likely to generalize outside of the training dataset, or that capture general cancer processes, etc. This plot just shows that those "general" features aren't necessarily more or less consistent than selecting features just based on variance (MAD), which isn't terribly surprising but is interesting to know.</p>

PR #52, opened 2022-09-08
F-statistic distribution heatmaps

jjc2718:
For my committee meeting presentation, Casey and I thought a heatmap showing the distribution of univariate f-statistics across cancer types would be useful. The idea is to show that there are some features that are strongly correlated with the labels in one cancer type and not others, and some features that are correlated across many cancer types.

EGFR is a good example of this - when we select by pan-cancer f-statistic (middle heatmap) we can see that most of the genes/features are strongly correlated in LGG and not as correlated in other cancer types, but when we select by median f-statistic (right heatmap) correlations tend to be more spread out across cancer types.

EGFR_heatmaps


arielah (approved):
This seems pretty straightforward and it looks good! Your in-file documentation is helpful.

PR #54, opened 2022-09-08
CCLE mutation prediction

jjc2718:
**PR description:**

This PR implements mutation status prediction (sample has/does not have a mutation in the given gene) for cell line data from CCLE, using gene expression. In general this is harder than it was for TCGA because there are about an order of magnitude fewer cell lines than there were tumor samples in TCGA (see #51 for some analysis of cancer types and mutated proportion of cell lines in the CCLE data).

Likely because of this, mutation prediction doesn't work as well on CCLE data as it did for TCGA, both in terms of overall performance and in terms of effectiveness of our feature selection method. Here are the results for EGFR, a gene that worked well on TCGA as shown in #48:

EGFR_all_summary

EGFR_non_carcinoma_summary

So we can see that there's very little difference between `pancan_f_test` and `median_f_test`, and also that most of the `delta_aupr` values are close to 0 suggesting that our models aren't doing much better than one with randomly permuted labels.

**Code changes:**

* `08_cell_line_prediction/run_ccle_mutation_prediction.py` to run experiments
* `pancancer_evaluation/data_models/ccle_data_model.py` and `pancancer_evaluation/utilities/ccle_data_utilities.py` to load the CCLE data using the same preprocessing pipeline we used for the TCGA data
* Some small edits to `pancancer_evaluation/utilities/data_utilities.py` to correctly load oncogene/TSG info
* `08_cell_line_prediction/plot_mutation_prediction_results.ipynb` to visualize results


08_cell_line_prediction/scripts/plot_fs_for_genes.sh
```diff
@@ -0,0 +1,35 @@
...
+)
+
+for gene in "${genes[@]}"; do
+
+    cmd="papermill plot_mutation_prediction_results.ipynb "
+    cmd+="${papermill_output_dir}/plot_mutation_prediction_results.run.ipynb "
+    cmd+="-k ${kernel_name} "
+    cmd+="-p gene ${gene}"
+    echo "Running: $cmd"
+    eval $cmd
```
ben-heil, line 31: I hadn't heard of papermill before, I assume it's better for running notebooks than `nbconvert --execute`?

ben-heil (approved):

08_cell_line_prediction/scripts/plot_fs_for_genes.sh
jjc2718: It's pretty much the same but it takes command line parameters, and `nbconvert --execute` doesn't support them AFAIK. I think officially Jupyter notebooks aren't really intended to take parameters like a function/Python module would, but I find it convenient if I want to run the same notebook for multiple genes or multiple cancer types, etc.

jjc2718:
<p>Good catch! I think I'll just remove it for now, I'm planning to make some tweaks to this in the coming weeks so any interpretation I add now is likely to change.</p>

---

 <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/54/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

PR #55, opened 2022-09-16
Ridge regression and other small tweaks

jjc2718:
This PR makes a few small changes to the cell line mutation prediction pipeline:

* We want to shuffle labels within cancer types, rather than across the whole dataset, for our baseline models
* Made the logic for single cancer/pancancer/all other cancers a bit more straightforward
* Added an option for ridge regression rather than elastic net (this tends to be a bit faster/easier to fit for large experiments)

We also ran a few experiments looking at results across different numbers of features selected for the cell line data. It doesn't really seem to change the results too much from what we saw before - the models all seem to sort of converge to the same performance value as the number of features gets higher, which makes sense since the feature selection method would have less of an influence if we're including a ton of features.

Examples:

TP53_ridge_non_carcinoma_summary

KRAS_ridge_non_carcinoma_summary


pancancer_evaluation/utilities/classify_utilities.py
```diff
@@ -819,6 +862,33 @@ def generate_param_grid(cv_results, fold_no=-1):
...
 
 
+def shuffle_by_cancer_type(y_df, seed):
+    y_copy_df = y_df.copy()
+    with temp_seed(seed):
+        for cancer_type in y_copy_df.DISEASE.unique():
+            is_cancer_type = (y_copy_df.DISEASE == cancer_type)
+            y_copy_df.loc[is_cancer_type, 'status'] = (
+                np.random.permutation(y_copy_df.loc[is_cancer_type, 'status'].values)
+            )
```
ben-heil, line 872: You can take or leave this suggestion; I initially thought `is_cancer_type` denoted whether the given disease was a cancer type. 
```suggestion
            cancer_type_indices = (y_copy_df.DISEASE == cancer_type)
            y_copy_df.loc[cancer_type_indices, 'status'] = (
                np.random.permutation(y_copy_df.loc[cancer_type_indices, 'status'].values)
            )
```

ben-heil (approved):

08_cell_line_prediction/plot_ridge_results.ipynb
```diff
@@ -0,0 +1,1215 @@
+{
```
ben-heil, line 1: <p>Is there a reason pancan f-test goes down as the number of features increases? Is it because it has a harder time selecting the best ones?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/55/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

jjc2718: <p>Some of it is probably noise (like how pancan f-test goes up and down in the second and third plots) but it is interesting that the best performers overall are the f-test methods with fewer features. This dataset is quite a bit smaller than TCGA (in terms of the number of samples, i.e. cell lines) so my guess is that all the methods probably just start overfitting, relative to the more aggressively feature selected models, when more features are included.</p><p> </p><p>I'd like to see if the same thing happens with the TCGA data; my guess is the inflection point will be higher (more features included) since there are more samples.</p>

Issue #53, opened 2022-09-08
Mutation prediction outstanding issues/updates


jjc2718:
- [ ] `use_pancancer` should be false for `all_other_cancers` in `process_data_for_gene()` (i.e. there shouldn't be unused dummy/one-hot variables)
- [ ] Label filtering should happen after we take the intersection of samples between gene expression and mutation (this will make the proportions in `08_cell_line_prediction/download_data.ipynb` match what we actually see when the scripts run)
- [ ] `tcga_utilities` should probably be renamed to something more general, or split
- [ ] CNV data for cell lines, in `ccle_data_model` `_generate_labels()`
- [ ] remove unknown/non-cancerous samples in `load_sample_info()`

PR #56, opened 2022-09-24
CCLE drug response classification, stratified by cancer type

jjc2718:
This is the first PR of (probably) several setting up drug response prediction on CCLE data as a use case for our feature selection methods. Here, we took the cell line classifications from [Iorio et al. 2016](https://doi.org/10.1016/j.cell.2016.06.017) into resistant/sensitive for a few drugs, and we're trying to predict them using gene expression data. For now we're stratifying CV folds by cancer type (so train/test sets have equal representation of the same cancer types), which should be the "easy" case compared to holding out entire cancer types.

In general, we don't get great performance for the six drugs we're looking for, with many AUPR values near 0 or only slightly better:

image

These are pretty similar to the AUPR values reported in https://arxiv.org/abs/2208.14822 (see Table 5: they're using multi-omics data so their results are slightly better than ours, but not by too much). So this is a slightly harder problem than we expected, and honestly we don't see much separation between feature selection methods with most of them performing comparably to random features.

We'll have to think about whether this is the right problem, or if regression on continuous drug response values (e.g. IC50 values) would be a better way to go. Our labels are pretty imbalanced here (the vast majority of cell lines are resistant to the vast majority of drugs) so that could be one issue.


08_cell_line_prediction/download_drug_data.ipynb
```diff
@@ -0,0 +1,1078 @@
+{
```
ben-heil, line 1: <p>What is the difference between <code>0</code> and <code>NaN</code> in  this matrix? Is <code>0</code> a known non-responder where  <code>NaN</code> doesn't have any data? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/56/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

08_cell_line_prediction/plot_stratified_drug_response.ipynb
```diff
@@ -0,0 +1,576 @@
+{
```
ben-heil, line 1: <p>Do you know what the baseline AUPR is for these? I know you said the classes are imbalanced so is it possible that a mean AUPR of ~.3 is actually better (or worse) than the baseline?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/56/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

ben-heil, line 1: <p>Probably can remove this cell :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/56/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

pancancer_evaluation/data_models/ccle_data_model.py
```diff
@@ -131,6 +169,42 @@ def _generate_labels(self, gene, classification, gene_dir):
...
+          # NA can occur in label column or COSMICID column
+          # in either case we want to drop those samples here
+          .dropna()
+          .astype({drug: int, 'COSMICID': int})
+          .set_index("SAMPLE_BARCODE")
+          .rename(columns={drug: 'status', 'cancer_type': 'DISEASE'})
+        )
+
+        # filter for cancer types without an extreme label imbalance
+        # TODO: intersect with X_df first?
```
ben-heil, line 196: Does this need to stay or did you figure it out?

ben-heil (approved):

08_cell_line_prediction/download_drug_data.ipynb
jjc2718: Yep, exactly - `NaN` are cell lines that weren't assayed for the relevant drug, `0` are cell lines that were assayed and were resistant/non-responders. I'll add a comment.

08_cell_line_prediction/plot_stratified_drug_response.ipynb
jjc2718: <p>Haha, will do!</p>

jjc2718: <p>I assume by "baseline AUPR" you mean something like <a href="https://stats.stackexchange.com/a/266989" target="_blank">what's described here</a>? That would definitely be useful to look at. My guess is that our "shuffled" models here are actually outperforming that baseline in most cases - since we're shuffling labels within each cancer type (i.e. maintaining label proportions within cancer types), I would imagine that just predicting a different mean value for each cancer type would outperform the overall label proportion across the dataset.</p><p> </p><p>I added this to my "outstanding issues" <a href="https://github.com/greenelab/pancancer-evaluation/issues/53" target="_blank">GitHub issue</a> and I'll try to add it soon. My scripts don't currently save information about the label proportion of the training set after preprocessing/filtering etc. but they probably should.</p>

PR #58, opened 2022-10-03
Drug response regression, stratified by cancer type

jjc2718:
The main goal of this PR is to build out regression (predicting continuous values) in addition to classification (predicting binary values) for drug response. I've applied it here to stratified cross-validation, as a way to make sure everything's working smoothly and we're generally doing better than our baselines, which we do seem to be.

Here's a plot comparing the true labels with shuffled labels across a few drugs, measuring performance using Spearman correlation between predictions and true labels:

image

So in most cases the blue boxes are considerably higher than the orange boxes, which is good! Next we'll try the same holdout experiments as before (liquid vs. solid, and single cancer type holdouts) using regression as well.

Major code changes:
* Moved model training code out of `pancancer_evaluation/utilities/classify_utilities.py` and created `pancancer_evaluation/prediction/classification.py` and `pancancer_evaluation/prediction/regression.py` for classification and regression model fitting respectively
* Add code to `08_cell_line_prediction/download_drug_data.ipynb` and `pancancer_evaluation/utilities/ccle_data_utilities.py` to process/load continuous labels (we're using log(IC50) values here which are provided by GDSC, [here's the Wiki page about IC50](https://en.wikipedia.org/wiki/IC50))
* Plot results in `08_cell_line_prediction/plot_stratified_drug_regression.ipynb`


08_cell_line_prediction/plot_stratified_drug_regression.ipynb
```diff
@@ -0,0 +1,771 @@
+{
```
ben-heil, line 1: <p>I guess I haven't seen the whole table, but is it weird that you have positive correlation in the training data and no/negative correlation in the test data? Is that just stratification noise?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/58/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

pancancer_evaluation/prediction/classification.py
```diff
@@ -0,0 +1,355 @@
...
+    results_append_list = [
+        gene,
+        holdout_cancer_type,
+        signal,
+        seed,
+        data_type,
+        fold_no,
+    ]
+
+    metrics_out_ = [results["auroc"], results["aupr"]] + results_append_list
```
ben-heil, line 284: Is there a reason for the trailing underscore here? Is there a `metrics_out` variable that you're avoiding a conflict with?

ben-heil (approved):

pancancer_evaluation/prediction/classification.py
jjc2718: No reason that I know of. I copied this part of the code from one of Greg's repos a long time ago, so I'm not sure exactly what he was going for. It looks like the trailing underscore in sklearn code is sometimes used for [variables that are only defined after the model is fit](https://stackoverflow.com/a/66205119), which I guess would apply here, but this function really only makes sense to call after the model is fit anyway so I don't think the trailing underscore adds much. I'll remove it here!

08_cell_line_prediction/plot_stratified_drug_regression.ipynb
jjc2718: <p>Yeah, I think in this case it's probably just stratification/cross-validation noise combined with overfitting, in the sense that the best model in terms of validation set performance fits the training data closely but doesn't perform as well on the test data. Ultimately predicting erlotinib sensitivity in this dataset seems to be a pretty hard problem, so getting better than random correlation is difficult for some of the cross-validation folds (although across all the folds/iterations we still end up with median correlations that are greater than 0).</p>

PR #59, opened 2022-10-06
Drug response regression, cancer type holdout experiments

jjc2718:
Similar to #57, we wanted to try holding out some cancer types entirely, either liquid vs. solid or individual cancer types in CCLE, using regression instead of classification. Results are somewhat similar in the sense that there's not much separation between feature selection methods and selecting random features seems hard to beat, although performance over the baseline is probably somewhat better suggesting that regression rather than classification is probably the way to go.

When we look at individual cancer types we often see pretty high variability between CV folds. For example, here are results for docetaxel response prediction:

image

For e.g. cervical cancer or sarcoma, the error bars reach from -1 (perfect negative correlation between predicted/true labels) to 1 (perfect positive correlation between predicted and true labels). My guess is that this is because there are only a small handful of samples (maybe 2 or 3) in some of the test sets, so Spearman correlations of -1 or 1 are pretty likely if you guess labels at random, whereas for the cancer types with more cell lines (breast, colon, lung etc) we see much less variance.

As a next step we're thinking about filtering to cancer types with a certain number of samples (maybe 10 or 15) assayed for the given drug, which may cut down on some of the variance/randomness although I doubt it will drastically change results.


08_cell_line_prediction/plot_drug_response_regression.ipynb
```diff
@@ -0,0 +1,1285 @@
+{
```
ben-heil, line 1: <p>Is there an issue with the way the data is displayed int the table? r^2 of -8.6 shouldn't be possible, but maybe there's a "e-1" somewhere that's getting truncated</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/59/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

ben-heil (commented):
Looks good! If the issue with the r^2 values is just a graphical one then consider this an approval. Otherwise I can rereview once it's fixed

08_cell_line_prediction/plot_drug_response_regression.ipynb
jjc2718: I think a very negative r^2 value (even considerably more negative than -1) is possible for a really overfit model on a held-out dataset. [This article](https://towardsdatascience.com/explaining-negative-r-squared-17894ca26321) seems like a pretty decent explanation of how negative r^2 values can arise on test data.

To check this, I adapted the script in that article a bit to generate a dataset with more features and fewer train samples (i.e. to encourage overfitting). They generated 200 data points (160 train/40 test) with 10 features, and I changed this to 50 data points (40 train/10 test) with 50 features. Here are the r^2 values that gave me:

```
Train R2: 1.000, Test R2: -20.088
Train R2: 1.000, Test R2: -1.973
Train R2: 1.000, Test R2: -7.360
Train R2: 1.000, Test R2: -3.315
Train R2: 1.000, Test R2: -6.338
Train R2: 1.000, Test R2: -4.848
Train R2: 1.000, Test R2: -2.584
Train R2: 1.000, Test R2: -3.115
Train R2: 1.000, Test R2: -4.141
Train R2: 1.000, Test R2: -2.071
Train R2: 1.000, Test R2: -1.332
Train R2: 1.000, Test R2: -0.031
Train R2: 1.000, Test R2: -1.385
Train R2: 1.000, Test R2: -4.247
Train R2: 1.000, Test R2: -1.182
Train R2: 1.000, Test R2: -10.096
Train R2: 1.000, Test R2: -0.779
Train R2: 1.000, Test R2: -4.884
Train R2: 1.000, Test R2: -8.114
Train R2: 1.000, Test R2: -14.131
```
This is fairly similar to what I'm seeing in our regression results - the very negative r^2 values tend to happen for cancer types with relatively few cell lines (e.g. neuroblastoma, lymphoma) on the test data. The train r^2 values for those drugs/cancer types tend to be considerably higher (close to 1), so my guess is that they're just really overfit.

I'm not sure how to avoid this, exactly. We've tried filtering to cancer types that have more than a given number of cell lines in the dataset (I think I used 15 cell lines as the threshold) and it cut down on some of the variance in the results, but not that much. The general trend of the results is still the same, though, that there's little difference between feature selection methods and it's hard to beat the random selection baseline.

ben-heil:
Ahhh well today I learned something about R^2

PR #60, opened 2022-11-16
Simulation of domain-specific data with shared signal

jjc2718:
Following on from previous feature selection/domain adaptation PRs like #48 and #59, we're thinking about implementing some additional methods for domain adaptation and domain generalization. It makes sense to validate them first on some simulated data that reasonably approximates different domains (i.e. different datasets or cancer types, etc), so this PR lays the groundwork for those simulations.

To simulate the data, we're using the generative model described in [this paper](http://proceedings.mlr.press/v119/piratla20a/piratla20a.pdf). The general idea is that there is a latent variable that's shared across domains, and latent variables that are specific to each domain, which are multiplied by the label (+ some randomly sampled noise) to generate data points. The hope is that our models can extract the shared signal and use it to predict across domains.

Generally, we see much better performance within simulated domains than across them with standard (non-domain aware) models, so that makes sense! Next we'll try adding some domain adaptation methods.

image
image


09_simulations/csd_simulations.py
```diff
@@ -0,0 +1,174 @@
...
+            xs = np.concatenate((xs, xs_i))
+        if ys is None:
+            ys = ys_i
+        else:
+            ys = np.concatenate((ys, ys_i))
+        
+    return xs, ys
+
+
+def simulate_csd_corr(n_domains, n_per_domain, p, k, corr_top=1., diag=None):
```
arielah, line 124: You could consider merging this and `simulate_csd_corr` with a parameter of whether or not to use correlated noise, since the rest of the code seems similar.

09_simulations/models.py
```diff
@@ -0,0 +1,187 @@
...
+from sklearn.pipeline import Pipeline
+
+def train_ridge(X_train,
+                y_train,
+                seed,
+                c_values=None,
+                n_folds=3,
+                max_iter=1000):
+
+    if c_values is None:
```
arielah, line 21: Why not just have the c_values set up as the default? i.e. 
```
def train_ridge(X_train,
                y_train,
                seed,
                c_values=[1e-6, 1e-5, 1e-4, 0.001, 0.01, 0.1, 1, 10, 100, 1000],
                n_folds=3,
                max_iter=1000):
```

09_simulations/models.py
```diff
@@ -0,0 +1,187 @@
...
+def train_mlp(X_train,
+              y_train,
+              params,
+              seed,
+              batch_size=50,
+              n_folds=3,
+              max_iter=1000,
+              search_n_iter=20):
+
+    import torch.optim
```
arielah, line 106: Is there a reason you're importing them here?

09_simulations/nbconverted/simulate_domains.py
```diff
@@ -0,0 +1,372 @@
...
+simulate_same_z = True
+correlated_noise = True
+
+
+# In[3]:
+
+
+if k is not None:
+    if simulate_with_csd:
+        if correlated_noise:
```
arielah, line 58: Seconding my thought that correlated noise should be a boolean you pass to a single `simulate_csd` function to save yourself some lines of code.

09_simulations/nbconverted/simulate_domains.py
```diff
@@ -0,0 +1,372 @@
...
+axarr[0].legend()
+axarr[1].set_title('UMAP projection of simulated data, colored by domain')
+axarr[1].set_xlabel('UMAP dimension 1')
+axarr[1].set_ylabel('UMAP dimension 2')
+axarr[1].legend()
+
+
+# ### Random train/test splitting
+# 
+# Just split the data randomly here, and fit some models to the split data. This gives us an idea of what baseline performance we should expect when we have access to training data from all domains.
```
arielah, line 128: For clarity in the first sentence, can you say explicitly that you're splitting data from all of the 4 domains?

arielah (approved):
This looks really interesting! I'm surprised that just adding in a covariate for what domain something's from improves the results so much. Excited to see what you find. I only have a couple of hair-splitting suggestions.

09_simulations/csd_simulations.py
```diff
@@ -0,0 +1,174 @@
+import numpy as np
+from scipy.stats import ortho_group
+
+def simulate_no_csd(n_domains, n_per_domain, p, noise_scale=1.):
+    """Simulate data from a latent variable model with multiple
+    domains, without a common component across domains.
+    """
+    xs, ys = None, None
+    z = np.random.normal(size=(n_domains, p))
+    betas = np.random.uniform(-1, 2, size=(n_domains,))
```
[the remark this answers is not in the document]
jjc2718: This is a detail from the paper - they say this range is to allow the betas to sometimes be negatively correlated for an individual domain, but on average across all domains the beta values will have a positive correlation with the labels. I'll add a similar note to the code.

09_simulations/models.py
jjc2718: I've generally done this with my neural network code because getting pytorch installed and configured can be a bit annoying. If I put the import at the top, then anyone who runs the code will need to have pytorch installed, but if I put it in this function it will only be necessary if the user actually wants to fit a NN.

09_simulations/csd_simulations.py
```diff
@@ -0,0 +1,174 @@
+import numpy as np
+from scipy.stats import ortho_group
+
+def simulate_no_csd(n_domains, n_per_domain, p, noise_scale=1.):
+    """Simulate data from a latent variable model with multiple
```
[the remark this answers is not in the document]
jjc2718: Sure, I can do that. The only reason I separated all these functions out was so I could figure out which parts of the generative model are important and what they each do, just by adding them gradually. (Ironically the most important/influential was probably the domain-correlated noise which was the last thing I added)

I think now that I understand how it all works a bit better I can probably remove most of those functions, but I'll do that in my next PR.

PR #61, opened 2022-11-23
Linear model with CSD loss layer - initial implementation

jjc2718:
This PR contains an initial implementation of the "common-specific decomposition" (CSD) method described in [this paper](http://proceedings.mlr.press/v119/piratla20a/piratla20a.pdf), and some simple comparisons with other models to sanity check that it's working.

The method from the paper ("`linear_csd`" in the plots) performs similarly to the other methods on a random train/test split (i.e. all simulated domains are in both the train and test set):

image

But when an entire domain is held out, `linear_csd` outperforms the linear model and random forest, and performs similarly to the 3-layer NN despite having far fewer parameters:

image

This is enough to convince me that it's probably implemented correctly, although I'll explore a wider range of simulation parameters and test cases in a future PR.

**Code changes:**

* Added new model specification/CSD loss function to `09_simulations/nn_models.py`
* Updated training/evaluation loop code in `09_simulations/models.py`
* Ran experiments and visualized results in `09_simulations/simulate_domains.py`


09_simulations/models.py
```diff
@@ -162,12 +173,105 @@ def train_mlp(X_train,
     return cv_pipeline
 
 
+def train_linear_csd(X_train,
+                     y_train,
+                     train_domains,
+                     seed,
+                     n_domains=None,
+                     params=None,
```
arielah, line 181: Could you call it something more descriptive than params? It's tripping me up that there are some parameters embedded within, but others like n_folds and batch_size are their own separate variable. You could try to embed them all or you could just rename it something like "learning_params"

09_simulations/models.py
```diff
@@ -162,12 +173,105 @@ def train_mlp(X_train,
...
+                     latent_dim=2,
```
arielah, line 182: There's also a latent_dim parameter within `params`, will this cause problems?

09_simulations/nn_models.py
```diff
@@ -1,6 +1,10 @@
+import numpy as np
+import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
+from skorch import NeuralNet
```
arielah, line 6: I'm sure it's not, but that package name just looks like a typo.

09_simulations/nbconverted/simulate_domains.py
```diff
@@ -39,35 +30,28 @@
...
-                              noise_scale=noise_scale,
-                              corr_top=corr_top,
-                              diag=diag)
-    else:
-        xs, ys = simulate_no_csd_large_z(n_domains, n_per_domain, p, k, noise_scale)
-elif simulate_same_z:
-    xs, ys = simulate_no_csd_same_z(n_domains, n_per_domain, p, noise_scale)
-else:
-    xs, ys = simulate_no_csd(n_domains, n_per_domain, p, noise_scale)
+xs, ys = simulate_csd(n_domains, n_per_domain, p, k,
```
arielah, line 50: This looks a lot cleaner, nice job

arielah (approved):
Looks good! Just need to figure out what's going on with your declared parameters.

09_simulations/nn_models.py
jjc2718: Yeah, I think it's supposed to be a merger/portmanteau between `sklearn` and `torch` (since it's an sklearn wrapper for training pytorch models) but it is a bit of a weird name. 🤷

09_simulations/models.py
jjc2718: Removed! You're right, the function argument wasn't being used.

PR #65, opened 2023-01-19
Start working on LASSO experiments across all genes

jjc2718:
Following on #64, this PR continues to build the infrastructure to run and analyze similar experiments with model size/LASSO penalty variation across multiple cancer genes. In this case, I ran the same experiments for TP53, EGFR, ATRX and CDKN2A (all common cancer drivers with diverse cancer types included in the training set).

The script at `02_cancer_type_classification/lasso_range_gene.ipynb` breaks down performance across cancer types for a single gene. Here's an example for CDKN2A:

image

This shows that for most cancer types, the lower LASSO parameters (less regularization/more features) perform at least as well as the higher LASSO parameters (more regularization/fewer features), even when the test cancer type is held out of the training set.

By contrast, the script at `02_cancer_type_classification/lasso_range_all.ipynb` breaks down performance across all the genes in the dataset:

image

This shows that across these 4 genes, there tends to be a positive correlation between the number of features in the model and generalization performance (i.e. in general, more features in the model => better generalization). The next step is to scale this up to the set of several hundred driver genes from my last paper, and find a clean way to analyze/visualize it (i.e. not a box plot for each of >100 genes...)


02_cancer_type_classification/nbconverted/lasso_range_all.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+coefs_perf_df = (nz_coefs_df
+    .rename(columns={'cancer_type': 'holdout_cancer_type'})
+    .merge(perf_df[perf_df.signal == 'signal'],
+           on=['gene', 'holdout_cancer_type', 'seed', 'fold', 'lasso_param'])
+    .drop(columns=['signal'])
+)
+
+coefs_perf_df.head()
```
arielah, line 89: I think it could be helpful along with this, since you can only see the rows for which experiment is `all_other_cancers`, adding in some sort of visualization of how many rows for each value of experiment there are. Does that make sense?

02_cancer_type_classification/nbconverted/lasso_range_all.py
```diff
@@ -0,0 +1,155 @@
...
+
+# In[7]:
+
+
+# look at correlation for each cancer type individually
+# positive correlation => more features, better performance
+corr_cancer_type_df = []
+nz_coefs_cutoff = coefs_perf_df.nz_coefs.quantile(q=0.1)
+
+if nz_coefs_cutoff is not None:
```
arielah, line 113: This is a confusing line to read, could you throw a comment on top to show what this line is doing?

02_cancer_type_classification/nbconverted/lasso_range_all.py
```diff
@@ -0,0 +1,155 @@
...
+sns.set({'figure.figsize': (12, 6)})
+
+ax = sns.boxplot(data=corr_cancer_type_df, x='gene', y='pearson_r')
+ax.axhline(0.0, linestyle='--', color='black')
+plt.title(f'Model size/performance correlations across cancer types, per gene (nonzero cutoff: {nz_coefs_cutoff})')
+plt.xlabel('Gene')
+plt.ylabel('Pearson correlation')
+
+
+# This box plot suggests that across these 4 genes, there tends to be a positive correlation between the number of features in the model and generalization performance (i.e. in general, more features in the model => better generalization).
```
arielah, line 155: Correlations can be tricky. Maybe plot 1-2 non aggregated to make sure it look like there is a true positive correlation between # of features and performance?

arielah: Just realized I should have looked at lasso_range_gene first. Look like you did this already.

02_cancer_type_classification/nbconverted/lasso_range_gene.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+# ### Get coefficient information for each lasso penalty
+
+# In[3]:
+
+
+nz_coefs_df = []
+
+# get pancancer coefs info for now
```
arielah, line 52: Do you want to change this in the future? Maybe throw a TODO on?

02_cancer_type_classification/nbconverted/lasso_range_gene.py
```diff
@@ -0,0 +1,291 @@
...
+    columns=['gene', 'cancer_type', 'lasso_param', 'seed', 'fold', 'nz_coefs']
+)
+nz_coefs_df = nz_coefs_df[nz_coefs_df.gene == plot_gene].copy()
+nz_coefs_df.head()
+
+
+# In[4]:
+
+
+genes = ['TP53']
```
arielah, line 75: What's the point of this if you're plotting ATRX?

02_cancer_type_classification/nbconverted/lasso_range_all.py
```diff
@@ -0,0 +1,155 @@
...
+# ### Get coefficient information for each lasso penalty
+
+# In[3]:
+
+
+nz_coefs_df = []
+
+# get pancancer coefs info for now
+for coef_info in au.generate_nonzero_coefficients_lasso_range(results_dir):
+    (gene,
```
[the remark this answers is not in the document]
jjc2718: The function has multiple return values, and the line you highlighted is just unpacking them into separate variables. It's the same as doing:
```python
for (gene, cancer_type, seed, lasso_param, coefs_list) in au.generate_nonzero_coefficients_lasso_range(results_dir):
```
but I think unpacking the return values separately looks a bit nicer than having the whole huge tuple in the `for` block.

02_cancer_type_classification/nbconverted/lasso_range_gene.py
jjc2718: Good catch! Removed.

02_cancer_type_classification/nbconverted/lasso_range_gene.py
jjc2718: Ahh, this comment isn't accurate anymore since I have the `training_dataset` variable in the previous cell which can be used to select the dataset now. I'll get rid of it.

02_cancer_type_classification/nbconverted/lasso_range_all.py
jjc2718: There's actually only one "experiment" here at the moment, it's one of `single_cancer`, `pancancer`, or `all_other_cancers` depending on what `training_dataset` is set to at the top of the notebook. I think I used that column in a previous notebook when I was combining data from all 3 "experiments", but it's not really necessary here, at least not right now.

I'll just drop the column from this dataframe since it's not being used.

PR #66, opened 2023-01-27
Run LASSO range experiments for Vogelstein gene set

jjc2718:
This PR expands #65 to all the genes in the [Vogelstein cancer gene set](https://github.com/greenelab/pancancer/blob/master/data/vogelstein_cancergenes.tsv), and modifies the analyses/visualizations from the last PR a bit to handle lots more genes.

In general, we see that for most genes, there are either positive or ~0 correlations between the number of features and generalization performance across cancer types, suggesting that if anything, models that include more features tend to generalize better. We're still exploring this and thinking of ways to summarize the results.

image

**Code changes:**

- Moved LASSO analysis scripts to `02_cancer_type_classification/lasso_range_analysis` directory
- Modified `02_cancer_type_classification/lasso_range_analysis/lasso_range_all.ipynb` to use different correlation methods, and to plot distributions for all genes
- Added `02_cancer_type_classification/lasso_range_analysis/lasso_corr_analysis.ipynb` to compare correlation methods (i.e. to identify genes with high CCC and low Pearson/Spearman, or vice-versa)

`02_cancer_type_classification/lasso_range_analysis/lasso_range_gene.ipynb` doesn't really need to be reviewed, it's mostly the same as the previous script that was moved.


02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_range_all.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+base_results_dir = os.path.join(
+    cfg.repo_root, '02_cancer_type_classification', 'results', 'lasso_range_valid'
+)
+
+training_dataset = 'all_other_cancers'
+results_dir = os.path.join(base_results_dir, training_dataset)
+
+# cutoff to filter out "dummy regressor" over-regularized models
```
arielah, line 36: Very helpful comment!

02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_range_all.py
```diff
@@ -0,0 +1,280 @@
...
+plt.title('Distribution of feature count across cancer types/folds')
+plt.xlabel('Number of nonzero features')
+
+# calculate quantile cutoff if included
+if quantile_cutoff is not None:
+    nz_coefs_cutoff = coefs_perf_df.nz_coefs.quantile(q=quantile_cutoff)
+    plt.gca().axvline(nz_coefs_cutoff, linestyle='--')
+    print('cutoff:', nz_coefs_cutoff)
+    
+coefs_perf_df.loc[coefs_perf_df.nz_coefs.sort_values()[:8].index, :]
```
arielah, line 123: Based on the comment above, shouldn't the rows with 0 and 1 nz_coefs (I'm assuming means nonzero coefficients) be thrown out?

02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_range_all.py
```diff
@@ -0,0 +1,280 @@
...
+with sns.plotting_context('notebook', font_scale=1.2):
+    plt.xlim(-1.0, 1.0)
+    sns.histplot(data=mean_corr_cancer_type_df, x=corr_column)
+    plt.axvline(0.0, linestyle=':', color='black')
+    plt.axvline(mean_corr_cancer_type_df[corr_column].mean(), linestyle='--', color='blue')
+    plt.title(f'Distribution of average {print_corr_name(correlation)} correlations between model size and performance, per cancer type', y=1.02)
+    plt.xlabel(f'Mean {print_corr_name(correlation)} correlation')
+    plt.ylabel('Cancer type count')
+
+print(f'Mean of mean correlations: {mean_corr_cancer_type_df[corr_column].mean():.4f}')
```
arielah, line 279: This graph really confused me at first. You're looking at the average correlation across all different genes that could be predicted for each cancer type? What do you make of these results? Also, couldn't that be biased since there are some gene/cancer type pairs you're skipping? It would really help to add a little contextualization of this graph below.

02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_corr_analysis.py
```diff
@@ -0,0 +1,86 @@
...
+           values='corr')
+)
+
+corr_wide_df.head()
+
+
+# In[23]:
+
+
+gene_ranks_df = (corr_wide_df
```
arielah, line 58: It's hard to track what how you're aggregating and ranking these. Maybe add a little explanation?

02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_corr_analysis.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+# In[25]:
+
+
+gene_ranks_df['ccc_spearman_diff'] = (
+    gene_ranks_df.ccc - gene_ranks_df.spearman
+).abs()
+
+gene_ranks_df.sort_values(by='ccc_spearman_diff', ascending=False).head(8)
```
arielah, line 85: This may get a little chaotic, but could you look at the absolute difference between the ccc and spearman values outside of the ranking system. Would be interesting to see if the same pairs stick out.

arielah (approved):
Interesting stuff! Just a couple points of clarification.

02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_range_all.py
jjc2718: Yeah, that happens in the next cell - here I wanted to visualize a few of the rows that are being thrown out, before they're actually thrown out.  I'll add a comment saying this.

02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_range_all.py
jjc2718: Good question! This is definitely confusing, I'll add some explanation.

Like you said, in this plot I'm first taking the mean correlation per gene/cancer type combination, then averaging over genes for each cancer type. And yes, this could be biased if some cancer types are frequently skipped or skipped in certain genes, but on average I think it gives a coarse-grained idea of which cancer types really benefit from larger models, and which benefit more from regularized/smaller models.

I'll also add the gene count for each cancer type, just to see if there are any that seem to be very over/under-represented.

PR #67, opened 2023-02-02
Microsatellite instability prediction across cancer types

jjc2718:
This PR implements microsatellite instability (MSI) status prediction across cancer types, as another prediction problem to study for the LASSO parameter range experiments I've been looking at. TCGA has MSI classifications for COAD, READ, STAD, and UCEC, so these experiments use those 4 cancer types.

In general, lower LASSO penalties (less regularization) tend to work better for this problem, both for performance on the training cancer types and for generalization to unseen cancer types.

image

The script at `10_msi_prediction/download_msi.ipynb` was already reviewed as part of the https://github.com/greenelab/mpmp repo, so you don't need to look too closely at it. The other scripts are new, but based heavily on existing scripts used for the mutation prediction experiments.


10_msi_prediction/nbconverted/lasso_range.py
```diff
@@ -0,0 +1,182 @@
...
+nz_coefs_df.lasso_param = nz_coefs_df.lasso_param.astype(float)
+nz_coefs_df.head()
+
+
+# In[4]:
+
+
+sns.set({'figure.figsize': (12, 5)})
+
+sns.boxplot(
```
arielah, line 76: Since the boxplots on either end of the lasso parameters are so hard to see, could you switch it so the lines of the boxplot are colored, rather than the fill which you can't see on the narrow boxes? Not sure how exactly you'd do that in Seaborn but I'm sure it's possible

10_msi_prediction/nbconverted/lasso_range.py
```diff
@@ -0,0 +1,182 @@
...
+    g = sns.relplot(
+        data=plot_df,
+        x='lasso_param', y=metric, hue='data_type',
+        hue_order=['train', 'cv', 'test'],
+        kind='line', col='holdout_cancer_type',
+        col_wrap=3, height=4, aspect=1.2
+    )
+    g.set(xscale='log', xlim=(0, 0.1))
+    # g.set_xticklabels(rotation=70)
+    plt.suptitle(f'LASSO parameter vs. number of nonzero coefficients, MSI prediction', y=1.05)
```
arielah, line 172: Same thing here, think this title is inaccurate

10_msi_prediction/run_msi_prediction.py
```diff
@@ -0,0 +1,192 @@
...
+)
+from pancancer_evaluation.utilities.classify_utilities import run_cv_cancer_type
+import pancancer_evaluation.utilities.data_utilities as du
+import pancancer_evaluation.utilities.file_utilities as fu
+
+def process_args():
+    p = argparse.ArgumentParser()
+    p.add_argument('--classify', action='store_true')
+    p.add_argument('--feature_selection',
+                   choices=['mad', 'pancan_f_test', 'median_f_test', 'random'],
```
arielah, line 27: I've never seen this used in an argparse before, cool!

pancancer_evaluation/utilities/analysis_utilities.py
```diff
@@ -396,6 +434,39 @@ def generate_nonzero_coefficients_lasso_range(results_dir, gene=None):
...
+    """
+    coefs = {}
+    all_features = None
+    for coefs_file in os.listdir(results_dir):
+        if coefs_file[0] == '.': continue
+        if 'signal' not in coefs_file: continue
+        if 'coefficients' not in coefs_file: continue
+        cancer_type = coefs_file.split('_')[1]
+        seed = int(coefs_file.split('_')[-4].replace('s', ''))
+        lasso_param = coefs_file.split('_')[-2].replace('c', '')
```
arielah, line 461: Are you using negative indexing here because there's an unknown number of intermediate columns?

arielah (approved):
Looks good! This is well documented as well.

10_msi_prediction/nbconverted/lasso_range.py
jjc2718: Good idea! I do think this is easier to follow.

image

10_msi_prediction/nbconverted/lasso_range.py
```diff
@@ -0,0 +1,182 @@
...
+
+if output_plots:
+    plt.savefig(output_plots_dir / f'msi_lasso_facets.png',
+                dpi=200, bbox_inches='tight')
+
+
+# A few takeaways:
+# 
+# * Interestingly, MSI prediction doesn't really seem to benefit from regularization. The models with a very small lasso parameter (basically a direct minimization of the log-loss with very little sparsity penalty) perform equally well or better than the regularized models, both within the training cancer types (the "cv" line) and for generalization to unseen cancer types (the "test" line).
+# * Generalizing to UCEC from the other cancer types seems to be the hardest. This isn't too surprising since it only occurs in women, which likely makes it more different from the other carcinomas than they are from each other. Maybe adding a sex covariate to the models would improve generalization in this case, since this could help encourage the model to focus on other female samples.
```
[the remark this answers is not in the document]
jjc2718: Me too! That will probably be part of the next PR, I'm going to tweak a few things with this analysis.

pancancer_evaluation/utilities/analysis_utilities.py
jjc2718: Yeah, basically. When I was doing the feature selection experiments a while ago, there were a few feature selection methods (e.g. `pancan_f_test`, `median_f_test`) that had underscores in them, so I switched to negative indexing since that made the number of intermediate columns from the beginning of the filename unknown.

Using underscores in the feature selection methods probably wasn't the best idea in retrospect (I could have used dashes or periods or something else) but I don't think I realized that until I had run a bunch of computationally expensive experiments so I didn't want to change.

I'll make a note of this in the code, although it's not that important for this set of experiments since I'm not varying the feature selection method at all.

PR #68, opened 2023-02-09
Sex prediction results + add sex covariate for MSI prediction

jjc2718:
This PR does two different things - unfortunately they ended up being a bit coupled so it would have been hard to separate into distinct PRs.

* Add a patient sex covariate for MSI prediction experiments (this didn't really end up changing anything results-wise, generalization to UCEC is still difficult)
* Implement patient sex prediction. The results of this were a bit interesting: we expected really small models to work the best (e.g. just including genes on the X/Y chromosomes). This doesn't really seem to be the case, although performance is good overall even on held-out cancer types.

image

The best-performing LASSO penalty parameters generally correspond to models having hundreds to thousands of features. I haven't looked yet at whether these features tend to be on the X and Y chromosomes, although I may do that in the future.


pancancer_evaluation/utilities/analysis_utilities.py
```diff
@@ -466,8 +466,8 @@ def generate_nonzero_coefficients_lasso_range_msi(results_dir):
         # use negative indexing since some feature selection methods
         # have underscores in them - indexing from the back of the
         # filename is safer/more consistent
-        seed = int(coefs_file.split('_')[-4].replace('s', ''))
-        lasso_param = coefs_file.split('_')[-2].replace('c', '')
+        seed = int(coefs_file.split('_')[-5].replace('s', ''))
```
arielah, line 469: Will this mean you have problems when you try to process files generated under the old system?

10_msi_prediction/nbconverted/lasso_range.py
```diff
@@ -173,37 +178,69 @@
...
+)
+
+print(plot_df.nz_quantile.unique())
+plot_df.head()
+
+
+# In[11]:
+
+
+# try with a float-valued x-axis
```
arielah, line 218: This comment appears to be in here twice. Do you need it in both places?

10_msi_prediction/run_msi_lasso_range.sh
```diff
@@ -3,44 +3,49 @@
...
-        cmd="python 10_msi_prediction/run_msi_prediction.py "
-        cmd+="--lasso_penalty $lasso_penalty "
-        cmd+="--num_features 16148 "
-        cmd+="--results_dir $RESULTS_DIR "
-        cmd+="--seed $seed "
-        cmd+="--training_samples pancancer "
-        cmd+="2>$ERRORS_DIR/errors_${seed}_${lasso_penalty}_pancancer.txt"
-        echo "Running: $cmd"
-        eval $cmd
+        # cmd="python 10_msi_prediction/run_msi_prediction.py "
```
arielah, line 32: Do you want to keep this commented code in?

11_sex_prediction/nbconverted/lasso_range.py
```diff
@@ -0,0 +1,265 @@
...
+sns.set_style('whitegrid')
+sns.histplot(plot_df.nz_coefs)
+
+quantiles_df = []
+
+for q in np.linspace(0.1, 0.9, 9):
+    quantiles_df.append([q, plot_df.nz_coefs.quantile(q)])
+    plt.gca().axvline(x=plot_df.nz_coefs.quantile(q), color='black', linestyle='--')
+    
+quantiles_df = pd.DataFrame(quantiles_df, columns=['quantile', 'value'])
```
arielah, line 222: This is a pretty rudimentary question, but what do you make of the extreme bimodal distribution of nonzero coefficients? Is the range of lasso parameters you're giving causing them to have almost all or almost no nonzero coefficients? Does that mean it might be too extreme of a range?

arielah: I'm also concerned if quantiles is the right approach here. The difference between quantile 3 and 4 is  ~8000 genes, whereas the difference between quantile 7 and 8 and ~40. I feel like using the quantiles as the scale for your boxplot is misrepresenting things. Feel free to push back though.

arielah (approved):
Looks good! These results are an ever-twisting trail.

pancancer_evaluation/utilities/analysis_utilities.py
jjc2718: I shouldn't - I ran a bash script to fix all the old ones. It was pretty simple so I didn't track it in this repo. If I missed any I think I'd rather just fix them than add an "if" statement to handle that case.

pancancer_evaluation/utilities/data_utilities.py
```diff
@@ -302,6 +302,61 @@ def load_purity(mut_burden_df,
...
+def load_sex_labels(mut_burden_df,
+                    sample_info_df,
+                    verbose=False):
+    """Load sex labels for use as target variable."""
+
+    if verbose:
+        print('Loading patient sex info...', file=sys.stderr)
+
+    clinical_df = (
+        pd.read_excel(cfg.clinical_data,
```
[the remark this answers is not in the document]
jjc2718: Haha, this was pretty simple but in general I try to avoid Excel wherever I can!

10_msi_prediction/nbconverted/lasso_range.py
jjc2718: Nope! Good catch.

10_msi_prediction/run_msi_lasso_range.sh
jjc2718: Nope, it should be fine to get rid of it.

11_sex_prediction/nbconverted/lasso_range.py
jjc2718: These are both great questions. I think since we're still somewhat in the "throw things at the wall and see what sticks" stage of this project I'm not too concerned quite yet with getting these details exactly right. For instance, I'm not totally sure if we'll go forward with the quantile plots or some other way of summarizing the results, so I'll think this through more if the "bins" are the type of visualization we end up preferring.

To answer the first question in slightly more detail, usually when I'm first doing parameter tuning I prefer a slightly larger range, and I think for this project in particular we want to make sure to include really simple models (almost no nonzero coefficients) and really complex un-regularized models (almost all nonzero coefficients) so I chose a range that would cover those cases. But I do think for the next analysis I do (coming soon!) it will be more important to make sure the "middle range" is more densely covered so that's a great point that I agree with.

I'll create an issue for both of these questions, since I do want to keep them in mind going forward.

PR #71, opened 2023-02-15
Rerun mutation prediction + compare optimizers

jjc2718:
In #68, in addition to what I listed in the PR description, I also tried running MSI prediction with a different sklearn interface/optimizer. Generally we've been running most experiments using [SGDClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html), which optimizes the logistic loss using stochastic gradient descent. Instead I tried using [LogisticRegression](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html) with an L1 penalty using the `liblinear` optimizer, which uses a coordinate descent algorithm that's supposed to converge quickly but can scale worse to datasets with many samples.

Since performance was generally better for MSI prediction with `LogisticRegression`, but not _that much_ better, in this PR I reran the mutation prediction experiments from #65 using `LogisticRegression`, and compared the results between the two optimizers in the notebooks `02_cancer_type_classification/lasso_range_analysis/compare_optimizers_all.ipynb` and `02_cancer_type_classification/lasso_range_analysis/compare_optimizers_gene.ipynb`.

In general, it does seem like the `liblinear` optimizer results in a better fit for almost every gene:

image

In this plot, each sample is a gene/cancer type combination, and a positive value means `liblinear` performed better than `sgd` for the best-performing LASSO parameters using each optimizer.


02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_range_gene.py
```diff
@@ -146,198 +148,139 @@
...
+    plt.suptitle(f'LASSO parameter vs. {metric.upper()}, {plot_gene}', y=1.01)
 
 if output_plots:
     plt.savefig(output_plots_dir / f'{plot_gene}_lasso_facets.png',
                 dpi=200, bbox_inches='tight')
 
 
 # ### Compare feature selection with performance
+# 
+# TODO: this needs to be documented
```
arielah, line 185: Reminder if you want to address this before merging

02_cancer_type_classification/lasso_range_analysis/nbconverted/compare_optimizers_all.py
```diff
@@ -0,0 +1,318 @@
...
+sgd_results_dir = os.path.join(sgd_base_results_dir, training_dataset)
+
+metric = 'aupr'
+
+# "cv" to plot performance on validation set (same cancer types as training)
+# "test" to plot performance on held-out cancer type (not in training set)
+data_type = 'cv'
+
+
+# ### Get coefficient information for each lasso penalty
```
arielah, line 53: I'm not sure where you're using this info? My best guess is the plot at the bottom but that's sorted by "ll_sgd_diff" which I assume is the AUPR difference.

02_cancer_type_classification/lasso_range_analysis/nbconverted/compare_optimizers_all.py
```diff
@@ -0,0 +1,318 @@
...
+    .groupby('gene')
+    .agg(np.mean)
+    .sort_values(by='ll_sgd_diff', ascending=False)
+).index.values
+
+with sns.plotting_context('notebook', font_scale=1.5):
+    ax = sns.boxplot(data=optimizer_diff_df, order=gene_order, x='gene', y='ll_sgd_diff')
+    ax.axhline(0.0, linestyle='--', color='grey')
+    plt.xticks(rotation=90)
+    plt.title(f'Top {data_type} set performance difference between liblinear and SGD optimizers, per gene', y=1.02)
```
arielah, line 313: Looks like you're looking at the top performing set (among all CV folds I think?) would you have different results if you compared all of them? Is it worth doing?

02_cancer_type_classification/lasso_range_analysis/nbconverted/compare_optimizers_gene.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+sns.boxplot(
+    data=ll_nz_coefs_df.sort_values(by=['cancer_type', 'lasso_param']),
+    x='cancer_type', y='nz_coefs', hue='lasso_param', ax=axarr[0]
+)
+color_boxplot_edges(axarr[0])
+axarr[0].set_title('liblinear')
+
+sns.boxplot(
```
arielah, line 147: Do you want the legend to be in the middle of the plot? It sometimes overlaps the data, but you can still see the main idea of the plot so your call.

arielah (approved):
Looks good! I think one chunk of code got copied over from compare_optimizers_gene to compare_optimizers_all that doesn't need to be there, but I could be wrong.

02_cancer_type_classification/lasso_range_analysis/nbconverted/compare_optimizers_all.py
jjc2718: You're right, it's not being used in this script - good catch. Loading it here for all genes takes a while, so probably a good idea to remove it unless it's really needed.

02_cancer_type_classification/lasso_range_analysis/nbconverted/compare_optimizers_all.py
jjc2718: Sorry, this isn't super clear - by "set" I mean holdout dataset, since I'm doing this both for "cv" which is the validation set with the same cancer types as the training data, and "test" which is the test/held out cancer type. Then the boxes are the median/IQR performance over all held-out cancer types and all CV folds for that gene (so I think I'm already doing what you're suggesting if I understand correctly).

The "top performing" part refers to the LASSO parameters - I'm not sure it would make a ton of sense to do a comparison over all LASSO parameters, since some of them perform really badly which would drag everything toward 0.

Maybe calling it a "dataset" instead of "set" would make this title a bit clearer. I'll also put the explanation above in the notebook for future reference.

PR #72, opened 2023-02-24
TCGA -> CCLE mutation prediction experiments

jjc2718:
Instead of using cancer types within TCGA as the holdout/test set, like we've been doing, we wanted to try using an independent dataset. Some of Greg's work suggested that predicting mutation status in cell lines using models trained on TCGA would probably work, so I'm training mutation status prediction models on TCGA and testing on CCLE in this PR, using varying LASSO penalties like before.

In general, most models perform better on the TCGA validation dataset or around the same on TCGA and CCLE, although weirdly there are a few that perform better on the CCLE test dataset despite not being trained on cell line data.

image

For selecting LASSO parameters, he general trend is somewhat similar to holding out cancer types, but a bit less pronounced. For most genes it is still best to select models based on CV performance, but the difference between that and a parsimonious model isn't as large as it was within TCGA:

image

Each data point here is a gene, positive differences indicate that the top model using CV performance is better than picking the smallest model within the top 25% of CV performance.


08_cell_line_prediction/nbconverted/lasso_range_all.py
```diff
@@ -0,0 +1,374 @@
+#!/usr/bin/env python
```
arielah, line 1: Since this file and lasso_range_all.py exist in several directories with seemingly minimal changes (at least between this one and the one in 02_cancer_type_classification), would it be possible to combine them? Maybe not with the way that the notebooks work, but there seems to be enough shared material here that it may be worth putting them out into a functions file so if you make changes, you only have to change it in one analysis.

08_cell_line_prediction/nbconverted/lasso_range_gene.py
```diff
@@ -0,0 +1,276 @@
...
+    patch.set_facecolor('None')
+
+    # each box has associated Line2D objects (to make the whiskers, fliers, etc.)
+    # loop over them here, and use the same color as above
+    for line in ax.lines[i * lines_per_boxplot: (i + 1) * lines_per_boxplot]:
+        line.set_color(col)
+        line.set_mfc(col)  # facecolor of fliers
+        line.set_mec(col)  # edgecolor of fliers
+
+plt.title(f'LASSO parameter vs. number of nonzero coefficients, {plot_gene}')
```
arielah, line 102: Is it weird that there's a huge jump between 10 and 50 and between 100 and 250, but almost none between 50 and 100? Am I overthinking this?

08_cell_line_prediction/nbconverted/lasso_range_all.py
```diff
@@ -0,0 +1,374 @@
...
+    .sort_values(by='cv_test_aupr_diff', ascending=False)
+).index.get_level_values(0).values
+
+with sns.plotting_context('notebook', font_scale=1.5):
+    ax = sns.boxplot(data=best_perf_df, order=gene_order,
+                     x='gene', y='cv_test_aupr_diff',
+                     palette='flare')
+    ax.axhline(0.0, linestyle='--', color='grey')
+    plt.xticks(rotation=90)
+    plt.title(f'Difference between TCGA and CCLE mutation prediction performance, by gene', y=1.02)
```
arielah, line 371: It would be nice to see both average performance and difference in a single plot, e.g. that TP53 isn't much better or worse on CCLE but that's because it scores well on both vs. a gene with moderate performance on one and dismal performance on another. Maybe a scatterplot with mean validation AUPR on one axis and difference in validation and test AUPR on the other axis? Or would that be too messy?

08_cell_line_prediction/nbconverted/threshold_cnv.py
```diff
@@ -0,0 +1,118 @@
...
+sns.set({'figure.figsize': (12, 6)})
+ratio_samples = np.random.choice(ccle_cnv_df.values.flatten(), size=100000)
+g = sns.histplot(ratio_samples, bins=50)
+g.set_yscale('log')
+
+
+# In[4]:
+
+
+# we're using both of these just as rough thresholds that seem to work well
```
arielah, line 53: Good note, I was just wondering about that

08_cell_line_prediction/nbconverted/lasso_range_all.py
arielah: On further reflection, a lot of the inner workings are put in other functions, so ignore me.

08_cell_line_prediction/scripts/run_tcga_ccle_lasso.sh
```diff
@@ -0,0 +1,64 @@
...
+  "50"
+  "100"
+  "250"
+  "500"
+  "750"
+  "1000"
+)
+
+genes=(
+  # "KRAS"
```
arielah, line 34: Do you want to put these back in or leave them out?

pancancer_evaluation/config.py
```diff
@@ -40,9 +41,10 @@
 test_genes = ['TP53', 'KRAS', 'ARID1A']
 
 # gene/classification combos for stratified CV model tests
-stratified_gene_info = [('TP53', 'TSG'),
-                        ('KRAS', 'Oncogene'),
-                        ('ARID1A', 'TSG')]
+# stratified_gene_info = [('TP53', 'TSG'),
```
arielah, line 44: Same thing, do you want to keep TP53 out of it?

arielah (approved):
Looks good to me!

08_cell_line_prediction/nbconverted/lasso_range_gene.py
jjc2718: I think it's just that for this particular gene the C = 50 and C = 100 models happen to be about the same size, so the performance plateaus between those parameters. This doesn't seem to happen for other genes, for instance here's RB1 where the slope is fairly constant:

image

So I think since it's not consistent between genes I'm just going to chalk it up to model fitting randomness, at least for the time being.

08_cell_line_prediction/scripts/run_tcga_ccle_lasso.sh
jjc2718: These I can put back in! Good catch.

08_cell_line_prediction/nbconverted/lasso_range_all.py
jjc2718: That's a good idea! I was planning to add a "filtering" step where I get rid of bad models (we came up with a way to do this in my last paper that I can just use here again), but I think it would be interesting to look at the scatter plot too.

PR #74, opened 2023-03-06
Clean up LASSO plots + add shuffled baselines

jjc2718:
Two small changes in this PR:

* Clean up/unify some of the plotting code between scripts
* For the TCGA -> CCLE experiments in #72, I tested filtering out "bad" models by comparing against the baseline with shuffled labels. The resulting best vs. smallest diff plot looks pretty similar to what we had before:

image

As compared to the previous results (with "zero difference" models filtered out):

image


02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_range_all.py
```diff
@@ -204,40 +204,52 @@ def get_top_and_smallest_diff(gene, cancer_type):
...
+# In[10]:
 
 
 sns.set({'figure.figsize': (8, 6)})
+sns.set_style('whitegrid')
 
 sns.histplot(all_top_smallest_diff_df.top_smallest_diff)
-plt.title('Differences between top and smallest LASSO parameter')
-plt.xlabel('top - smallest')
+plt.title('Differences between "best" and "smallest good" LASSO parameter')
```
arielah, line 226: Do you define anywhere what "smallest good" is?

arielah (approved):
Looks good to me!

02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_range_all.py
jjc2718: Nope, not in the notebooks - I'll add some documentation.

PR #75, opened 2023-03-16
CCLE -> TCGA prediction experiments

jjc2718:
In #72 we looked at generalization of mutation prediction models from TCGA data to CCLE, in this PR we want to do the reverse - training models on CCLE data and evaluating them on TCGA.

In general, this seems to be a more difficult problem, which is what we expected to see. In this plot, genes above 0 performed better on the CCLE training data than on the TCGA test set, and vice-versa for genes below 0.

image

This is fairly different from the plot we saw in #72 where most genes were closer to 0. The "best model" vs. "smallest good model" analysis results are fairly similar to the TCGA -> CCLE experiments though, with not much difference between models for most genes and a slight shift toward "best" models, although maybe less pronounced.


02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_range_gene.py
```diff
@@ -317,7 +317,8 @@ def add_smallest_vline(data, **kws):
     ]
     legend_labels = ['"best"', '"smallest good"']
     l = ax.legend(legend_handles, legend_labels, title='Model choice',
-                  loc='lower left', bbox_to_anchor=(2.28, 1.3))
+                  # loc='lower left', bbox_to_anchor=(2.28, 1.3))
```
arielah, line 320: Do you want to get rid of the commented code?

arielah (approved):
Looks good!

02_cancer_type_classification/lasso_range_analysis/nbconverted/lasso_range_gene.py
jjc2718: I think I want to keep it, but I'll add an explanation for why it's there

PR #76, opened 2023-03-27
TCGA -> CCLE neural network experiments

jjc2718:
To follow up on our LASSO experiments, we wanted to try training neural networks, and see if generalization trends looked like the ones we saw when varying the LASSO parameter. We started by using training epochs in a similar way, with models trained for fewer epochs being considered more "regularized" and models trained for longer less so.

We didn't really see too much of the regularization effect we were looking for though - the models seem to train pretty quickly, and training/validation loss plateaus:

image

We're going to try poking at this result in a few different ways next, including trying dropout as the regularization axis, and varying/fixing the learning rate to see if different learning rates lead to more distinct regularization/early stopping effects.


pancancer_evaluation/utilities/classify_utilities.py
```diff
@@ -770,26 +832,37 @@ def shuffle_by_cancer_type(y_df, seed):
 def apply_model_params(train_model,
                        ridge=False,
                        lasso=False,
-                       lasso_penalty=None):
+                       lasso_penalty=None,
+                       model='lr'):
```
arielah, line 836: Do you make a comment of which values are accepted for `model`? Looks like it's just lr vs. mlp, but it could be good to have a comment explaining that lr encompasses ridge and lasso.

08_cell_line_prediction/nbconverted/nn_learning_curves.py
```diff
@@ -0,0 +1,75 @@
...
+# In[11]:
+
+
+sns.set_style('whitegrid')
+
+sns.lineplot(data=lc_df, x='epoch', y='value', hue='dataset')
+plt.xlabel('Epoch')
+plt.ylabel('AUPR')
+plt.ylim(0, 1.1)
+plt.title(f'Learning curve for {plot_gene}, averaged over folds')
```
arielah, line 55: Have you looked to see if the AUPR that it maxes out at matches the AUPR you see with logistic regression, both for the CV and test sets? You'd obviously expect to do about as well for a given gene with either model but my intuition says it's not a guarantee.

arielah (approved):
Looks good! Fairly straightforward analysis. Looks like your models are too good at training, would it be worth trying a harder problem like CCLE -> TCGA to see if you can get more of an observable effect?

pancancer_evaluation/utilities/classify_utilities.py
jjc2718: Good idea! I'll add a comment.

jjc2718:
> Looks like your models are too good at training, would it be worth trying a harder problem like CCLE -> TCGA to see if you can get more of an observable effect?

Yeah, that might be worth it. I think we're going to start by looking at other types of regularization like dropout and weight decay, but I'll keep this in mind as a possibility.

PR #77, opened 2023-04-03
Neural network parameter exploration experiments

jjc2718:
Following up on #76, since early stopping didn't really seem to work as a form of regularization, we wanted to explore some different ways to regularize neural networks that we could use to study generalization.

We tried varying the hidden layer size, the amount of weight decay, and the amount of dropout. For the most part, they look similar to the results for LASSO when there's a large amount of regularization, but they don't overfit in the same way that the LASSO models did when there's a small amount of regularization. Here are the curves for KRAS (we also looked at EGFR and the results are similar):

image
image
image


08_cell_line_prediction/nn_diagnostics/nbconverted/nn_dropout_range.py
```diff
@@ -0,0 +1,158 @@
...
+
+# In[3]:
+
+
+dropout_dfs = {}
+
+for dropout_file in glob.glob(
+    os.path.join(results_dir,
+                 plot_gene,
+                 f'{plot_gene}_signal_mad_s42_n{num_genes}_d*_classify_learning_curves.tsv.gz')
```
arielah, line 42: I'm assuming s42 refers to the seed, but if it's hardcoded, do you ever planning on trying with a different seed?

08_cell_line_prediction/nn_diagnostics/nbconverted/nn_dropout_range.py
```diff
@@ -0,0 +1,158 @@
...
+)
+                      
+print(perf_df.shape)
+perf_df.head()
+
+
+# In[8]:
+
+
+# same plot as before but with the "best"/"smallest" parameters marked
```
arielah, line 130: It doesn't look like the "best"/"smallest" parameters are marked here, is the comment outdated?

08_cell_line_prediction/nn_diagnostics/nbconverted/nn_hsize_range.py
```diff
@@ -0,0 +1,191 @@
...
+    dataset_labels = ['TCGA (train)', 'TCGA (holdout)', 'CCLE'] 
+    ax.legend(legend_handles, dataset_labels, title='Dataset')
+    sns.move_legend(g, "upper left", bbox_to_anchor=(1.01, 1))
+    plt.title(f'Hidden layer size vs. AUPR, {plot_gene}', y=1.025)
+
+
+# In[9]:
+
+
+# same plot as before but with the "best"/"smallest" parameters marked
```
arielah, line 164: Same thing here, I think the code got copied and changed but the comment wasn't. Looks like it's the same as above but with confidence intervals?

08_cell_line_prediction/nn_diagnostics/nbconverted/nn_lr_range.py
```diff
@@ -0,0 +1,105 @@
...
+            sns.lineplot(data=lc_df[lc_df.fold == fold],
+                         x='epoch', y='value', hue='dataset',
+                         legend=False, ax=ax)
+    ax.set_xlabel('Epoch')
+    ax.set_ylabel('AUPR')
+    ax.set_ylim(0, 1.1)
+    ax.set_title(f'Learning curve for {plot_gene}, lr={lr}, for each split')
+
+plt.tight_layout()
+
```
arielah, line 105: Any reason you don't do the final plot for this notebook?

08_cell_line_prediction/run_tcga_ccle_nn.py
```diff
@@ -57,20 +57,45 @@ def process_args():
     p.add_argument('--seed', type=int, default=cfg.default_seed)
     p.add_argument('--shuffle_labels', action='store_true')
     p.add_argument('--verbose', action='store_true')
+
+    # hyperparameter arguments
+    # if these are included they will override the random search ranges
+    # (in pancancer_evaluation/prediction/classification.py), instead providing
+    # a fixed value for the relevant hyperparameter
+    params = p.add_argument_group('params',
```
arielah, line 65: I didn't know you could have a group of arguments like this. Cool!

arielah (approved):
Looks good to me! Just a couple of tiny things.

08_cell_line_prediction/nn_diagnostics/nbconverted/nn_dropout_range.py
jjc2718: I'll add it as a parameter to the script(s)!

08_cell_line_prediction/run_tcga_ccle_nn.py
jjc2718: Yeah, it's a cool feature of argparse! It helps if you have a ton of arguments (which is often the case with NNs/complex ML models and their parameters).

PR #78, opened 2023-04-10
Cancer type holdout performance analysis

jjc2718:
For the holdout cancer type analyses (e.g. #66), we wanted to look at which cancer types tend to perform well/poorly as the holdout set.

In the box plot, positive values mean the classifiers performed better on the training data than on the holdout cancer type, and vice-versa. The results largely make sense: TGCT and SARC are non-carcinomas so it's not surprising that generalization was poor, THCA only has classifiers for 2 genes and one is very undersampled, etc.

image


arielah (approved):
Looks good to me!

PR #79, opened 2023-05-03
Optimizer comparison on stratified cross-validation data

jjc2718:
In order to separate the conclusion of the optimizer comparison into a separate small paper, we wanted to try it using stratified cross-validation (instead of holding out an entire cancer type or dataset); these changes enable that.

The results are mostly the same as we saw in #71: e.g. for KRAS:

image

And overall, liblinear performance is better than SGD for the best-performing model:

image


01_stratified_classification/nbconverted/lasso_range_all_optimizers.py
```diff
@@ -0,0 +1,393 @@
...
+            [gene, cancer_type, lasso_param, seed, fold_no, len(coefs)]
+        )
+        
+ll_nz_coefs_df = pd.DataFrame(
+    ll_nz_coefs_df,
+    columns=['gene', 'signal', 'lasso_param', 'seed', 'fold', 'nz_coefs']
+)
+ll_nz_coefs_df.lasso_param = ll_nz_coefs_df.lasso_param.astype(float)
+
+print(ll_nz_coefs_df.gene.unique())
```
arielah, line 71: Are these pre-selected by you or are these the actual non-zero genes that were picked by the trained model? If the latter, it's pretty impressive that all the "usual suspects" are there.

01_stratified_classification/nbconverted/lasso_range_all_optimizers.py
```diff
@@ -0,0 +1,393 @@
...
+ll_perf_df = au.load_prediction_results_lasso_range(ll_results_dir,
+                                                    'stratified')
+ll_perf_df = (ll_perf_df
+    .drop(columns=['holdout_cancer_type', 'experiment'])
+    .copy()
+)
+ll_perf_df.lasso_param = ll_perf_df.lasso_param.astype(float)
+
+print(ll_perf_df.shape)
+print(ll_perf_df.gene.unique())
```
arielah, line 116: Do you need to print this in every code block?

01_stratified_classification/nbconverted/lasso_range_all_optimizers.py
arielah: Nope, it's the Vogelstein genes, I'm an idiot.

01_stratified_classification/nbconverted/lasso_range_all_optimizers.py
```diff
@@ -0,0 +1,393 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# ### LASSO parameter range experiments, summary across all genes
```
arielah, line 4: Maybe a quick word here about what liblinear and SGD (stochastic gradient descent, I'm assuming?) are and why you care about the differences?

01_stratified_classification/nbconverted/lasso_range_gene_optimizers.py
```diff
@@ -0,0 +1,791 @@
...
+# In[11]:
+
+
+sns.set({'figure.figsize': (10, 5)})
+sns.set_style('ticks')
+sns.histplot(perf_coefs_df.nz_coefs)
+plt.title(f'Nonzero coefficient distribution for {plot_gene}')
+plt.xlabel('Number of nonzero coefficients')
+
+linear_bins_df = []
```
arielah, line 281: This list isn't getting used anywhere, and I'm not sure what the linear bins are anyway. They don't seem to match up with the bins being auto-generated for the histogram?

01_stratified_classification/nbconverted/lasso_range_gene_optimizers.py
```diff
@@ -0,0 +1,791 @@
...
+    [perf_coefs_df.optimizer == 'SGD']
+).loc[:, ['lasso_param', 'nz_linear_bin']]
+
+(sgd_param_bin_map
+   .drop_duplicates()
+   .groupby('lasso_param')
+)['nz_linear_bin'].apply(list)
+
+
+# In[15]:
```
arielah, line 348: This whole section needs some explanation. What are you trying to see with these plots?

01_stratified_classification/nbconverted/lasso_range_gene_optimizers.py
```diff
@@ -0,0 +1,791 @@
...
+sgd_plot_df = (
+    perf_coefs_df[(perf_coefs_df.optimizer == 'sgd')]
+      .sort_values(by=['nz_linear_bin'])
+      .reset_index(drop=True)
+)
+
+plot_df = pd.concat((ll_plot_df, sgd_plot_df))
+
+with sns.plotting_context('notebook', font_scale=1.6):
+    g = sns.relplot(
```
arielah, line 561: Okay, now I see that it's so you could have a standardized way to compare with the same directionality of penalty vs. number of coefficients. Probably need to add more documentation about that for future readers.

01_stratified_classification/nbconverted/lasso_range_gene_optimizers.py
```diff
@@ -0,0 +1,791 @@
...
+
+# In[27]:
+
+
+ll_top_lasso_param = ll_mean_perf_df.iloc[0, :].lasso_param
+sgd_top_lasso_param = sgd_mean_perf_df.iloc[0, :].lasso_param
+print(ll_top_lasso_param, sgd_top_lasso_param)
+
+
+# In[28]:
```
arielah, line 705: For readability, the rest of the file should probably be its own subsection about comparing coefficient size, with some explanation of why you're interested in it.

01_stratified_classification/scripts/run_lasso_compare_vogelstein.sbatch
```diff
@@ -0,0 +1,105 @@
...
+read_genes_from_file $merged_filename
+
+gene=${genes[${SLURM_ARRAY_TASK_ID}]}
+
+LL_RESULTS_DIR=./01_stratified_classification/results/optimizer_compare_ll
+SGD_RESULTS_DIR=./01_stratified_classification/results/optimizer_compare_sgd
+
+# number of features to "preselect" to by mean absolute deviation
+# -1 == no preselection
+MAD_PRESELECT=16042
```
arielah, line 39: What's the logic behind this number?

arielah (approved):
The code all looks good, this analysis got pretty confusing in places so I recommended some spots for additional documentation for future readers.

01_stratified_classification/nbconverted/lasso_range_all_optimizers.py
jjc2718: It is! I'll make a note of this though, "all genes" is pretty ambiguous

01_stratified_classification/nbconverted/lasso_range_all_optimizers.py
jjc2718: I generally do this to catch bugs with how the data is being read or generated - I've found that it helps to eyeball what the dataset looks like in terms of shape and columns included, etc. I guess in this case it's most important that I generated data with all of the same genes for SGD and liblinear, since I wouldn't expect them to have exactly the same number of rows due to differences, so maybe I can explicitly check that and make a note of it.

01_stratified_classification/nbconverted/lasso_range_gene_optimizers.py
jjc2718: I guess it makes sense to have them be the same, I was just using the default number of bins

01_stratified_classification/scripts/run_lasso_compare_vogelstein.sbatch
jjc2718: That's the number of genes in the gene set, that we have gene expression values for. My scripts mostly default to 8000 (we were using this for my past paper and other work, partially for computational reasons and partially because that's what Greg's mutation classification paper used, so we wanted our models to be comparable to those).

For this project I think it's a bit more important to use all the genes, since we're interested in model sparsity the background set for that should probably be all the genes in the gene set rather than a subset of them. I'll make a note of this in this script.

PR #82, opened 2023-06-09
Tweaks to figures for latest optimizer paper draft

jjc2718:
No huge changes here, just tweaks to figure formatting for the latest draft of our paper. I added some code in `01_stratified_classification/lasso_range_gene_optimizers.ipynb` to plot the loss functions for each gene/optimizer, but that should be the only substantial change.


01_stratified_classification/nbconverted/lasso_range_gene_learning_rate.py
```diff
@@ -189,15 +214,99 @@
...
+    for ax in g.axes.flat:
+        ax.set_ylabel('Loss value', visible=True)
+        ax.tick_params(labelleft=True)
+    g.set_titles('Learning rate schedule: {col_name}')
+    sns.move_legend(g, "center", bbox_to_anchor=[1.075, 0.5], frameon=True)
+    g._legend.set_title('Loss component')
+     
+    plt.suptitle(f'LASSO parameter vs. training loss, {plot_gene}', y=1.01)
+
+plt.tight_layout(w_pad=8, h_pad=5)
```
arielah, line 308: What's the significance of the vertical orange line? Also why is it partially shaded on the optimal one?

arielah (approved):
Looks good! The final plots look funky on first glance, so maybe make sure to explain what's going on with the orange stuff in the figure caption?

01_stratified_classification/nbconverted/lasso_range_gene_learning_rate.py
jjc2718: Those are places in the plots where the loss function is 0 (and on the optimal one the shading is a confidence interval since some CV folds are 0 and others aren't). I think it makes more sense visually to just have it be a very small number (and explain it in the figure caption) instead of log(0) = -infinity, so I'll make that change.

PR #85, opened 2023-08-03
Plot results for neural network experiments, across all genes

jjc2718:
Also cleaning up some other figures for the paper draft. This PR touches a lot of files but the changes aren't that substantial, most of them are just cosmetic.


02_cancer_type_classification/lasso_range_analysis/lasso_range_all.ipynb
```diff
@@ -17,6 +17,7 @@
    "source": [
```
nrosed, line 17: <p>Is it expected that theres a big difference in the number of genes in comparison to the previous run? (51480 now, 35520 before)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/85/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

nrosed, line 17: <p>is it also expected that there is a shift in the distribution from most being near 0 to not most being near 15000?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/85/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

nrosed, line 17: <p>is it strange that the test_auroc is so low for CIC-UCEC?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/85/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

02_cancer_type_classification/lasso_range_analysis/lasso_range_gene_cancer_type.ipynb
```diff
@@ -0,0 +1,752 @@
+{
```
nrosed: <p>Its very interesting that the stratified holdout and the test peak very close to each other.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/85/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

nrosed:
Looks good to me; the main thing that would help future readers are some main- and sub-section headers for some of the scripts.

Other comments were just conceptual questions I had on some of your plots. I might have just missed the intention of the plots, so it ignore them if they don't make any sense.

nrosed (approved):

02_cancer_type_classification/lasso_range_analysis/lasso_range_all.ipynb
jjc2718: <p>The "shape" of this dataframe is the total number of models, i.e. one for each combination of gene/cancer type/LASSO parameter. In the most recent run we covered a larger range of LASSO parameters, so it makes sense that the dataframe is considerably larger.</p><p> </p><p>There are also two genes that are included now that weren't before, ALK and CIC. I slightly changed how the filtering for hypermutated samples works which results in these genes being included, so this is expected.</p>

jjc2718: <p>That's the result of the larger range of LASSO parameters I described before. It is expected, and I would argue that it makes more sense for the model comparisons since we're covering a more comprehensive range of "overfit" or non-sparse models now.</p>

jjc2718: <p>It's something we've observed before for some genes/cancer types. We think it's because some mutations aren't all that correlated with gene expression in some cancer types (i.e. they're not drivers in that cancer type, or they don't occur frequently enough to train a good model, etc) so they end up overfitting, which is what we see here with good CV performance and bad performance on UCEC samples.</p><p> </p><p>Currently, we're just including/excluding cancer types based on mutation frequency/proportion. We could do something smarter but I think for this study where we have hundreds of genes/cancer types, it's probably fine to have the filter be pretty coarse-grained. A while back we tried filtering for models that outperform a baseline where the gene expression data is permuted, and it didn't affect the overall model selection results that much.</p>

02_cancer_type_classification/lasso_range_analysis/lasso_range_gene_cancer_type.ipynb
jjc2718: <p>It was interesting to us too! I think it's pretty typical though, at least for this mutation prediction problem across genes and cancer types. That's one of the main takeaways from our paper, that cross-validation performance is generally pretty predictive of generalization performance, and model simplicity may not be as important.</p>

jjc2718:
<p>Yeah, this is a "just for me" plot. The ones that show the train/valid/test breakdown are what we'll show in the paper, so I spent more time making them pretty.</p>

---

 <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/85/discussion/#comment-1666134378'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

jjc2718:
<p>Yeah, I'm not sure why there are so many genes that perform better on CCLE. It could be something technical or dataset-specific like what you're describing, or it could be something biological like the cell line data is just cleaner/more well-behaved in these cases than the tumor samples from TCGA. If I had more time maybe I'd try to detangle the two, but I have to start writing my thesis at some point, haha.</p>

---

 <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/85/discussion/#comment-1666134468'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

jjc2718:
<p>I'll add them! These scripts started as just a model diagnostic thing, but we are using a few of these figures in the supplement so I'll add some documentation.</p>

---

 <a href='https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/85/discussion/#comment-1666147145'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

jjc2718:
> View / edit / reply to [this conversation](https://app.reviewnb.com/greenelab/pancancer-evaluation/pull/85/discussion/#comment-1666134517) on ReviewNB
> 
> _nrosed commented on 2023-08-04T20:24:40Z_ ----------------------------------------------------------------
> 
> I don't fully understand this plot. Why is test error always at 0.6 even though the cv and train errors are going down?

For whatever reason, these models seem to saturate really fast rather than improving slowly across epochs. Here's a comparison of learning rates I did a while ago for KRAS mutation prediction:

image

For the lower learning rates, you can see that CV/test performance improves a bit more gradually, but the resulting CV and test performance after 200 epochs ends up being about the same as it is for the higher learning rates (other than the obviously bad ones like 0.01).

In the plots you're looking at for hidden layer size, I was doing a grid search (I think over the same range shown here) and choosing the best learning rate. So what that ends up picking happens to be one of the models that saturates really fast, at least for KRAS; I haven't looked too much at other genes.

We could add some kind of baseline, but since what we ultimately care about is the "best vs. smallest good" model selection comparison I don't think it's that important in this case. Like you mentioned, because of the considerable variability between hidden layer sizes I'm fairly confident that the model is learning something, and our goal isn't really to find the absolute best performing NN model here, just a reasonable one that allows us to think about model complexity by comparing models with different performances across many genes.

